PR #22, opened 2025-10-18
Add a 'my_templates' directory for pptx skill

jasonkholden:
Make it easy for a user to drop their favorite templates in a directory for use with the pptx skill

image


jasonkholden:
This might be a little heavy-handed for a general case but thought I'd put this out there for users managing just a few relatively static templates that they use

PR #28, opened 2025-10-18
Internal CSS properties (alphabetical order)

Placewith5s:
- Changed the order of the internal CSS properties (a-z)

**Other:**
- Fixed a CSS compatibility issue


nsa644 (commented):
good

PR #41, opened 2025-10-19
Basic changes for step 3

AKTechWiz:


veyselaksin:
💡 Suggestion for Improvement
Concurrency for Performance: Currently, URLs are checked sequentially. This will be slow when dealing with a large number of URLs. Implementing concurrency using concurrent.futures.ThreadPoolExecutor to check URLs simultaneously would be a significant performance enhancement.

PR #43, opened 2025-10-20
Clarify the need to run `/plugin` to see the plugin menu options

jxmallett:
I was confused by where the menu options were because I've not used `/plugin` before.


Molret (approved):
good point

PR #87, opened 2025-11-08
Add nested table extraction to docx skill

belumume:
## Summary

Adds nested table extraction pattern to the "Reading and analyzing content" section.

## Problem

python-docx's `cell.text` property doesn't traverse nested tables, causing cells with forms, checklists, or complex layouts to appear empty.

## Solution

Use `cell.tables` property to detect and extract nested table content.

## Use Cases

- Government forms (checkbox grids)
- Evaluation forms (rating scales)
- Surveys (multiple choice options)
- Business checklists (option lists)

## Changes

- Added `### Nested table extraction` subsection under "Reading and analyzing content"
- Includes detection pattern, extraction function, and usage example
- Covers basic extraction and recursive extraction for deep nesting
- 73 lines, self-contained

## Testing

- Tested on documents with nested tables (forms, evaluations, checklists)
- Code examples verified and functional

## References

- Stack Overflow: https://stackoverflow.com/questions/64334884/
- python-docx docs: https://python-docx.readthedocs.io/

### DIFF
--- a/document-skills/docx/SKILL.md
+++ b/document-skills/docx/SKILL.md
@@ -51,6 +51,152 @@ You need raw XML access for: comments, complex formatting, document structure, e
 * `word/media/` - Embedded images and media files
 * Tracked changes use `<w:ins>` (insertions) and `<w:del>` (deletions) tags
 
+
+## Advanced Pattern: Nested Table Content Extraction
+
+### Problem
+
+python-docx's `cell.text` property only extracts direct paragraph text. It does **not** traverse nested tables within cells.
+
+**Symptom:**
+```python
+cell.text  # Returns: '' or '
+'
+# But cell visually contains checkbox options, form fields, etc.
+```
+
+This occurs in:
+- Government forms (checkbox grids)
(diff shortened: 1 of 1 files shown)


belumume:
Closing this as obsolete rather than rebasing it.

The change is 70 lines of `python-docx` guidance: `cell.text` does not traverse nested tables, so cells containing them read as empty, with a helper that walks `cell.tables`. That was accurate against the skill as it stood when I opened this.

The skill has since been rewritten and no longer uses `python-docx` at all. It creates with `docx-js` and edits by unpacking and modifying `word/document.xml` directly, and the section this patch anchored to is gone. Rebasing would drop `python-docx` API guidance into a document that teaches neither the library nor that workflow, which would be worse than leaving the gap.

Checked before closing: the nested-table behaviour itself has not been documented elsewhere in the skill, so this is being closed because the approach no longer fits rather than because it was superseded.

If `python-docx` returns to that path the change is worth revisiting, and this can be reopened.

PR #143, opened 2025-12-14
Add guidance to avoid whitespace reformatting in docx/ooxml skil

cdxiaodong:
These changes ensure that documents processed by the docx skill will remain compatible with Microsoft Word by preserving the original XML formatting without adding unnecessary whitespace.


sbguangha:
pnpm is not supported, it originally only supported npm

PR #155, opened 2025-12-19
Fix links in agent skills specification

saqibameen:
The current links are broken in the spec, this PR fixes that


klazuka:
Thanks. I just merged the same fix in #159

PR #156, opened 2025-12-19
Correct small filename typos in agent-skills-spec.md

tmustier:
2 typos where underscores used instead of dashes:
- skill_authoring.md -> `skill-authoring.md`
- skill_client_integration.md -> `skill-client-integration.md`

This broke the links inside `agent-skills-spec.md`


klazuka:
thanks, I just merged the same fix in #159

PR #282, opened 2026-01-25
docs(webapp-testing): add requirements.txt for dependencies

brandonhawi:
# Why?

Every time I install and run this skill on a new machine, Claude Code writes scripts, runs them, THEN realizes that it may not yet have Playwright. This is a minimal change to hint to Claude Code (or whatever skills supporting agent) that playwright is needed. Follows the `requirements.txt` pattern as per `slack-gif-creator`, though personally I would have rather altered the skill to just advise `uv`.


brandonhawi:
@klazuka any chance I can get a look at this?

PR #284, opened 2026-01-26
Fix(skill-creator): specify utf-8 encoding when reading SKILL.md

drfccv:
Explicitly specifying utf-8 encoding when reading SKILL.md ensures consistent behavior across different operating systems and environments. It prevents potential UnicodeDecodeError exceptions that may occur if the system default encoding is not utf-8, especially on Windows. This change improves the script's robustness and compatibility when handling non-ASCII characters in SKILL.md files.
image


pperez:
i saw some other issues related to encoding on windows, `init_skill.py`, same thing in the `package_skill.py`

PR #359, opened 2026-02-09
Add YAML quoting requirement for skill descriptions to prevent parser errors

alexandre-machado:
This pull request adds an important clarification to the documentation for writing YAML frontmatter in `SKILL.md`. The update emphasizes the requirement to wrap the `description` value in double quotes if it contains YAML special characters, preventing parsing errors.

Documentation improvements:

* Added a note that the `description` field in YAML frontmatter must always be wrapped in double quotes when it contains special characters (such as colons, commas, brackets, quotes, or starts with a dash) to avoid YAML parsing errors.

image


laneser:
This PR solved my problem too! 👍

PR #465, opened 2026-02-25
chore: export latest skills

kencheeto:
Exports the latest skill builds, including updated skill-creator and docx skills.


klazuka (approved):

zackZhu1:
@kencheeto  Is there a reason why init_skill.py was removed? I found this script very useful, especially since it generates boilerplate directory and code for a skill

kencheeto:
> @kencheeto[Slackkenshiro](slack://user?team=T01HP5FJC81&id=U096Q4HHM9N) Is there a reason why init_skill.py was removed? I found this script very useful, especially since it generates boilerplate directory and code for a skill

We were increasingly finding that it wasn't necessary for Claude to be able to create valid skills (esp given the validation logic already in place). Glad you found it useful though - you should be able to retrieve it here if you'd like to keep it https://github.com/anthropics/skills/blob/1ed29a03dc852d30fa6ef2ca53a67dc2c2c2c563/skills/skill-creator/scripts/init_skill.py

zackZhu1:
Thanks! @kencheeto 
I looked through the changes in this PR, it looks like the skill-creator skill can now launch agents to evaluate the quality of the skill, which is awesome.

PR #487, opened 2026-03-01
frontend-design: Add `text-transform: uppercase` rule to typography guidelines

retlehs:
Adds a rule to the Typography section of the frontend-design skill clarifying that all-caps text should never be written directly in markup, and should instead use `text-transform: uppercase` in CSS.

This is both a semantic/accessibility best practice (screen readers may read uppercase letters individually) and a maintainability concern — keeping presentation in CSS rather than content.


DevPintu:
Thank you for adding this guideline — it's a valuable clarification for ensuring semantic markup and improving accessibility.
Happy to help review.

PR #521, opened 2026-03-05
Add record-knowledge skill

LevNas:
## Summary

Claude Code loses all context between sessions. A workaround you discovered yesterday must be rediscovered today. This skill lets Claude Code record such knowledge as tagged Markdown entries in `.claude/knowledge/entries/`, so it persists across sessions and is available to the whole team via Git.

## What's included

- `SKILL.md` with recording procedures, tag management, and entry linking
- Template file for `.claude/knowledge/CLAUDE.md` (tag registry + search commands)

## How it works

1. Claude Code autonomously records knowledge discovered during work sessions
2. Entries use YAML frontmatter with title, author, status, and tags
3. Tags are deduplicated via a similarity check against the tag registry
4. Related entries are cross-linked with bidirectional `see:` links
5. Entries are mutable — edit in place, git tracks history

## Installation

```
/install anthropics/skills:record-knowledge
```

### DIFF
--- a/skills/record-knowledge/SKILL.md
+++ b/skills/record-knowledge/SKILL.md
@@ -1,0 +1,134 @@
+---
+name: record-knowledge
+description: >-
+  Record tacit knowledge — quirks, pitfalls, dependencies, decisions, root causes — as tagged
+  Markdown entries in `.claude/knowledge/entries/`. Use this skill whenever discoveries are made
+  during work, when the user shares undocumented system behavior, or at plan completion to capture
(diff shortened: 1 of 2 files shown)


LevNas:
Updated the skill with a new section based on production usage:

- **"When to Reference"**: Guides Claude Code to search existing knowledge entries at session start and check for updates during progress updates. This was missing from the initial submission and has been valuable in practice for maintaining continuity across sessions.

Issue #532, opened 2026-03-06
skill-creator: description optimizer requires ANTHROPIC_API_KEY, unusable for enterprise/SSO users


betfanatics-codywang:
## Context

The skill-creator's description optimization loop (`scripts/run_loop.py`, `scripts/improve_description.py`, `scripts/run_eval.py`) requires `ANTHROPIC_API_KEY` to function. Most Claude Code users authenticate via enterprise SSO or Anthropic's managed license — they don't have (and can't easily obtain) a raw API key. This makes the description optimizer effectively unusable for the majority of Claude Code's user base.

Checked existing issues — #518 (eval loop rough edges) and #490 (workspace location) cover other skill-creator issues but not this one.

### 1. `run_loop.py` hardcodes `anthropic.Anthropic()` client

```python
# run_loop.py line 78
client = anthropic.Anthropic()
```

This crashes immediately if `ANTHROPIC_API_KEY` is not set, even though the user is fully authenticated with Claude Code via SSO.

### 2. `improve_description.py` uses the Anthropic SDK directly

The improvement step calls `client.messages.create(...)` to generate better descriptions. There's no fallback to `claude -p`, which would use the user's existing Claude Code authentication.

### 3. `run_eval.py` strips auth-related env vars too aggressively

```python
# run_eval.py line 83
env = {k: v for k, v in os.environ.items() if k != "CLAUDECODE"}
```

While this only strips `CLAUDECODE`, the broader pattern of env manipulation for subprocess `claude -p` calls is fragile. When users attempt workarounds that strip `CLAUDE*` vars more broadly (a natural instinct to avoid nesting issues), it breaks authentication entirely — producing `"Not logged in · Please run /login"` errors.

### 4. Real skills compete with eval temp commands

`run_single_query` creates a temporary command in `.claude/commands/` and checks if Claude invokes it. However, if a real skill in `.claude/skills/` has the same or similar description, Claude preferentially triggers the real skill instead of the temp command. The eval only counts triggers of the temp command, producing false-negative 0% trigger rates even when the description is excellent.

## Impact

- Users on enterprise Claude Code licenses (the majority of Claude Code users) cannot use the description optimization workflow at all
- The skill-creator SKILL.md prominently features description optimization as a key capability, but it silently fails for most users
- Workarounds require significant monkey-patching of the upstream scripts (we built a ~400-line wrapper to make it work)

## Suggested fix

Add a `claude -p` fallback path when `ANTHROPIC_API_KEY` is not set:

[...]

PR #534, opened 2026-03-06
fix(skill-creator): add claude -p CLI fallback for description optimizer

betfanatics-codywang:
## Summary

The skill-creator's description optimizer (`run_loop.py`, `run_eval.py`, `improve_description.py`) currently hard-requires `ANTHROPIC_API_KEY` for the improvement step and has a bug in env var stripping that breaks authentication for the eval step. This makes the optimizer unusable for the many users authenticated via Claude Code's enterprise SSO, Bedrock backend, or other methods that don't set `ANTHROPIC_API_KEY`.

**Root cause**: `run_eval.py` strips only `CLAUDECODE` from the environment when spawning `claude -p` subprocesses. However, `CLAUDE_CODE_ENTRYPOINT` is also a nesting guard that must be stripped, and — critically — no other `CLAUDE*` vars should be stripped because they contain auth tokens needed by the subprocess.

## Changes

1. **Fix env var stripping** (`run_eval.py`): Only strip `CLAUDECODE` and `CLAUDE_CODE_ENTRYPOINT` (the two nesting guards). The previous code stripped only `CLAUDECODE`, leaving `CLAUDE_CODE_ENTRYPOINT` which prevents nested invocations.

2. **Add CLI fallback for improve step** (`improve_description.py`): When `ANTHROPIC_API_KEY` is not set, fall back to `claude -p --output-format text` for description improvement calls. This uses whatever auth method the user has configured in Claude Code.

3. **Add skill-hiding during eval** (`run_eval.py`): Temporarily rename `SKILL.md` files to `.eval-backup` during trigger evaluation so real skills don't compete with eval command files (which causes false negatives). Restored on exit via context manager.

4. **Add `--setting-sources project,local`** (`run_eval.py`): Exclude user-scope plugins during eval for cleaner, more deterministic results.

5. **Graceful degradation** (`run_loop.py`): Auto-detect API key availability and route to the appropriate backend with a verbose log message.

## Motivation

When running the description optimizer from within a Claude Code session (the primary use case for the skill-creator skill), users who are authenticated via enterprise SSO do not have `ANTHROPIC_API_KEY` set. The eval step fails silently (0% trigger rate due to "Not logged in" errors in subprocesses), and the improve step crashes with an SDK authentication error.

This PR makes the optimizer work out-of-the-box for all Claude Code authentication methods.

Fixes #532

## Test plan

- [ ] Run optimizer with `ANTHROPIC_API_KEY` set → should use SDK path (existing behavior, unchanged)
- [ ] Run optimizer without `ANTHROPIC_API_KEY` but with Claude Code SSO auth → should use CLI fallback, triggers evaluate correctly
- [ ] Verify SKILL.md files are hidden during eval and restored after (including on error)
- [ ] Verify `--setting-sources project,local` excludes user-scope plugins from eval

### DIFF
--- a/skills/skill-creator/scripts/improve_description.py
+++ b/skills/skill-creator/scripts/improve_description.py
@@ -2,22 +2,94 @@
 """Improve a skill description based on eval results.
 
 Takes eval results (from run_eval.py) and generates an improved description
-using Claude with extended thinking.
+using Claude with extended thinking. Falls back to ``claude -p`` CLI when
+the Anthropic Python SDK is not available (no ANTHROPIC_API_KEY).
 """
 
 import argparse
 import json
+import os
 import re
+import subprocess
 import sys
 from pathlib import Path
 
-import anthropic
+try:
+    import anthropic
+
+    _HAS_SDK = bool(os.environ.get("ANTHROPIC_API_KEY"))
+except ImportError:
+    anthropic = None  # type: ignore[assignment]
+    _HAS_SDK = False
 
 from scripts.utils import parse_skill_md
 
+# Environment variables that prevent nested claude -p invocations.
+_NESTING_GUARD_VARS = ("CLAUDECODE", "CLAUDE_CODE_ENTRYPOINT")
+
+
+def _improve_via_cli(
+    prompt: str,
+    model: str,
+    log_dir: Path | None = None,
+    iteration: int | None = None,
+) -> str:
+    """Fallback: use ``claude -p`` CLI when the Anthropic SDK is unavailable.
+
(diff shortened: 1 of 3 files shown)


betfanatics-codywang:
Closing — this was superseded by #547 (commit b0cbd3df), which took the CLI-only approach on `main`, making the SDK-with-fallback pattern in this PR redundant. The `run_eval.py` improvements (nesting guard fix, `_hide_skills`, `--setting-sources`) may still be worth pulling into a separate PR if they aren't already on `main`.

Issue #189, opened 2025-12-30
document-skills and example-skills plugins install identical content, causing duplicate skills


chuggies510:
## Description

When installing both `document-skills` and `example-skills` plugins from `anthropic-agent-skills`, they contain **identical skills**, causing duplicates in Claude Code's context window.

## Expected Behavior

According to the repo README:
- `document-skills` should contain only the production document skills (docx, pdf, pptx, xlsx)
- `example-skills` should contain the open-source educational examples (algorithmic-art, canvas-design, etc.)

## Actual Behavior

Both plugins install **all 17 skills** from the `skills/` directory:

```
=== document-skills ===
algorithmic-art, brand-guidelines, canvas-design, doc-coauthoring, docx,
frontend-design, internal-comms, mcp-builder, pdf, pptx, skill-creator,
slack-gif-creator, theme-factory, web-artifacts-builder, webapp-testing, xlsx

=== example-skills ===
(identical list)
```

Both use the same commit hash in the cache (`69c0b1a06741`).

## Impact

- **~50k wasted tokens** on duplicate skill definitions
- Skills like `pptx` (6.3k tokens) appear twice in `/context`
- Reduces available context for actual work

## Steps to Reproduce

1. Install Claude Code
2. Run: `claude plugin install document-skills@anthropic-agent-skills`
3. Run: `claude plugin install example-skills@anthropic-agent-skills`
4. Run `/context` in Claude Code
5. Observe duplicate skills in the "Plugin" section

## Environment

- Claude Code **v2.0.76**
- macOS 15.6.1
- Both plugins installed at user scope

## Suggested Fix

The plugin packaging should respect the intended separation, or the marketplace should only expose one unified plugin that contains all skills without duplication.

## Workaround

Uninstall one of the duplicate plugins:
```bash
claude plugin uninstall example-skills@anthropic-agent-skills
```

ajmccaus:
I can second this issue. I am running the same version of Claude Code in Windows 11. I did some testing with my own local plugin repo and discovered the root cause, so this is a pattern to apply to any flat monorepo where you want to separate the skills into different installations. I also confirmed on this repo by adding using /plugin marketplace add anthropics/skills, modifying ~/.claude/plugins/marketplaces/anthropic-agent-skills/.claude-plugin/marketplace.json as below, and loading it as normal using /plugin install document-skills@anthropic-agent-skills.

# Solution for GitHub Issue #189: Duplicate Skills

## Root Cause

[...]

PR #551, opened 2026-03-06
fix: separate plugin source paths to prevent duplicate skill installation

drewburchfield:
## Summary

I was checking whether my setup would auto-pull updates from this marketplace and noticed that all three plugins in `marketplace.json` have their `source` set to `"./"`. This causes Claude Code to copy the entire repo into each plugin's cache, so every plugin ends up with all 17 skills regardless of what the `skills` array declares.

This has been broken since the marketplace was first published. PR #129 made it worse by consolidating all skills into a single `skills/` directory while keeping `"source": "./"` on every plugin.

If the intent is to expose three separate installable plugins (which makes sense), the `source` path needs to point to a dedicated subdirectory per plugin. That's the pattern used by `claude-plugins-official` and every other marketplace I checked.

## What changed

- Moved skills into per-plugin subdirectories:
  - `document-skills/skills/` (xlsx, docx, pptx, pdf)
  - `example-skills/skills/` (algorithmic-art, brand-guidelines, canvas-design, etc.)
  - `claude-api-plugin/skills/` (claude-api)
- Updated `marketplace.json` source paths from `"./"` to `"./document-skills"`, `"./example-skills"`, and `"./claude-api-plugin"`
- Updated README links to match new paths
- Removed the now-empty top-level `skills/` directory

## Impact

Before: installing all 3 plugins gives you 51 skill entries (17 x 3). ~56 MB of redundant cached files. Every skill tripled in the context window.

After: installing all 3 plugins gives you 17 skill entries (4 + 12 + 1). Each plugin only contains its declared skills.

Fixes #189

cc @klazuka @kencheeto @zack-anthropic


kimkimsh:
Bumping this — issue still reproduces in Claude Code as of 2026-05-06, affects 3 plugins (`claude-api` included), and #189 has accumulated 8+ confirmations. Would appreciate maintainer review.

drewburchfield:
Closing this. The behavior I originally reported no longer reproduces, and what's left doesn't justify the size of this change.

I re-tested against current Claude Code, installing the marketplace from a clean clone of `main` into an isolated `CLAUDE_CONFIG_DIR`:

- `document-skills` reports `Skills (4)`, `example-skills` reports `Skills (12)`, `claude-api` reports `Skills (1)`.
- I planted an undeclared skill directly into an installed plugin's cache. It did not load.
- `strict: true` vs `strict: false` made no difference.

So the `skills` array is authoritative at runtime, exactly as documented, and the context-window duplication I described in March is fixed client-side. My original summary above is wrong as of today. Leaving it for the record.

Two things `source: "./"` still causes, for anyone who lands here later:

**1. Cache duplication.** Each plugin copies the entire repo into its own cache. Installing all three produces a 33 MB plugin cache instead of 11 MB, and each plugin's cache contains `spec/`, `template/`, and the other plugins' skills.

**2. Redistribution of the source-available skills.** The `example-skills` and `claude-api` plugin caches each contain full copies of `docx`, `pdf`, `pptx`, and `xlsx`, which the README describes as "source-available, not open source." Installing either plugin writes them to disk.

Neither is fixable from the manifest alone. Both need a per-plugin `source` root, which is what this PR does. But that means moving ~400 files, and it collides with active `claude-api` work every time that tree changes. Not worth carrying as an outside PR against a repo that's moving this fast.

Leaving #189 open, since (1) and (2) are still real. The rebased version of this change is on my fork if a maintainer ever wants to pick it up.

PR #674, opened 2026-03-18
skill-creator: add design-phase guidance from internal skill-building lessons

abe238:
This PR adds design-phase guidance to the skill-creator skill based on the patterns documented in @thariqshihipar's article ["Lessons from Building Claude Code: How We Use Skills"](https://www.linkedin.com/pulse/lessons-from-building-claude-code-how-we-use-skills-thariq-shihipar-iclmc).

## The problem

The skill-creator is strong on the eval/iterate loop and description optimization, but gives less guidance during the design phase. Users jumping in to create a skill get great tooling for testing and benchmarking, but not much help thinking about *what makes a good skill* before they start writing one.

## What this adds

Seven new sections, all integrated into the existing structure:

1. **Skill type taxonomy** (9 categories) added to the Capture Intent phase. Helps users identify what kind of skill they are building so they can pick the right patterns early.

2. **"Focus on What Claude Does Not Already Know"** as a writing principle. The most common waste in skills is restating things Claude would do anyway. This section coaches users to focus on the non-obvious stuff: internal conventions, edge cases, domain-specific patterns.

3. **Gotchas section** as a first-class skill component. The article calls this the "highest-signal content in any skill." Now explicitly recommended in the writing guide.

4. **Setup and user config pattern** (`config.json`). Many skills need user-specific context (which Slack channel, which repo, credentials). This documents the check-on-first-run pattern so skills feel polished instead of asking the same questions every time.

5. **Memory and data persistence guidance**. Skills can store state across runs, but data in the skill directory gets wiped on upgrade. This section documents the problem and points to `${CLAUDE_PLUGIN_DATA}` as the stable alternative.

6. **On-demand hooks as a design pattern**. The frontmatter reference mentions hooks exist, but this section explains *when and why* to use them, with concrete examples like `/careful` and `/freeze`.

7. **Distribution guidance**. Repo-checked vs marketplace, and the context cost tradeoff that matters at scale.

## What this does not change

The eval loop, benchmarking, description optimization, and all environment-specific instructions (Claude.ai, Cowork) are untouched. This is purely additive.

## Why

I have been building and maintaining skills across multiple agent frameworks for a few months now. These gaps kept coming up. When Thariq published the article documenting the same patterns from Anthropic's internal usage, it felt like the right time to contribute them back upstream so others can skip the trial and error.

cc @thariqshihipar

### DIFF
--- a/skills/skill-creator/SKILL.md
+++ b/skills/skill-creator/SKILL.md
@@ -53,6 +53,24 @@ Start by understanding the user's intent. The current conversation might already
 3. What's the expected output format?
 4. Should we set up test cases to verify the skill works? Skills with objectively verifiable outputs (file transforms, data extraction, code generation, fixed workflow steps) benefit from test cases. Skills with subjective outputs (writing style, art) often don't need them. Suggest the appropriate default based on the skill type, but let the user decide.
 
+### Identify the Skill Type
+
+Before drafting, it helps to know what category the skill falls into. Most useful skills cluster into one of these types:
+
+| Type | What it does |
+|------|-------------|
+| **Library/API Reference** | Explains how to correctly use a library, CLI, or SDK. Focuses on edge cases, footguns, and gotchas that Claude would not know from training data alone. |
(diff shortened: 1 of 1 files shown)


abe238:
@ThariqS a gift for you here.

PR #867, opened 2026-04-04
Add git validation stop hook to prevent premature PR creation

tedted23:
## Summary
This PR adds a git repository validation hook that prevents pull request creation when there are uncommitted changes, untracked files, or unpushed commits. The hook is designed to be used as a "Stop" hook in Claude Code settings to catch common mistakes before initiating a PR workflow.

## Key Changes
- **stop-hook-git-check.sh**: New bash script that validates git repository state by checking for:
  - Uncommitted changes (both staged and unstaged)
  - Untracked files that should be committed
  - Unpushed commits on the current branch or against the default branch
  - Recursion prevention to avoid infinite loops
  - Graceful handling when not in a git repository

- **settings.json**: New Claude Code settings configuration that registers the git validation script as a Stop hook, ensuring it runs before PR-related operations

## Implementation Details
- The script reads JSON input from stdin and checks the `stop_hook_active` flag to prevent recursion
- Uses `git diff`, `git ls-files`, and `git rev-list` to comprehensively validate repository state
- Handles both cases where a remote branch exists and where it doesn't (comparing against origin/HEAD)
- Returns exit code 2 with descriptive error messages when validation fails, preventing PR creation
- Includes a reminder message to only create PRs when explicitly requested by the user

https://claude.ai/code/session_019TkrB2s4QCATL3sykRtxwD

### DIFF
--- a/backup/settings.json
+++ b/backup/settings.json
@@ -1,0 +1,19 @@
+{
+    "$schema": "https://json.schemastore.org/claude-code-settings.json",
+    "hooks": {
+        "Stop": [
+            {
+                "matcher": "",
+                "hooks": [
+                    {
+                        "type": "command",
+                        "command": "~/.claude/stop-hook-git-check.sh"
+                    }
+                ]
+            }
+        ]
+    },
+    "permissions": {
+        "allow": ["Skill"]
+    }
+}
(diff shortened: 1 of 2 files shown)


tedted23 (commented):
GG

tedted23:
afdd

PR #897, opened 2026-04-09
Add proper front-matter to SKILL.md for claude-api

quuu:
otherwise it's not a proper skill, and can't be installed or read properly by agents

PR #898, opened 2026-04-09
fix: claude-api skill frontmatter yaml format

allenzhou101:
A previous [PR](https://github.com/anthropics/skills/pull/897)￼ added skill frontmatter, but the YAML is malformed, preventing installation via `npx skills` and causing it to render incorrectly on GitHub. This fixes the issue by adding quotes around the description.

Before:
https://github.com/anthropics/skills/blob/main/skills/claude-api/SKILL.md

2026-04-09 at 11 52 15@2x

After:
https://github.com/allenzhou101/skills/blob/a/skill-yaml-rendering/skills/claude-api/SKILL.md
2026-04-09 at 12 00 41@2x

### DIFF
--- a/skills/claude-api/SKILL.md
+++ b/skills/claude-api/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: claude-api
-description: Build, debug, and optimize Claude API / Anthropic SDK apps. Apps built with this skill should include prompt caching. TRIGGER when: code imports anthropic/@anthropic-ai/sdk; user asks to use the Claude API, Anthropic SDKs, or Managed Agents (/v1/agents, /v1/sessions, /v1/environments). DO NOT TRIGGER when: code imports `openai`/other AI SDK, general programming, or ML/data-science tasks.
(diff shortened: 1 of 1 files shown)


Medenor:
I came for this specific problem, I support the fix 👍🏻

skills/claude-api/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: claude-api
-description: Build, debug, and optimize Claude API / Anthropic SDK apps. Apps built with this skill should include prompt caching. TRIGGER when: code imports anthropic/@anthropic-ai/sdk; user asks to use the Claude API, Anthropic SDKs, or Managed Agents (/v1/agents, /v1/sessions, /v1/environments). DO NOT TRIGGER when: code imports `openai`/other AI SDK, general programming, or ML/data-science tasks.
+description: "Build, debug, and optimize Claude API / Anthropic SDK apps. Apps built with this skill should include prompt caching. TRIGGER when: code imports anthropic/@anthropic-ai/sdk; user asks to use the Claude API, Anthropic SDKs, or Managed Agents (/v1/agents, /v1/sessions, /v1/environments). DO NOT TRIGGER when: code imports `openai`/other AI SDK, general programming, or ML/data-science tasks."
```
dev-khalid, line 3: Yaml formatmatter should be okay without this double quotation.

allenzhou101: YAML allows it but GitHub and `npx skills` parsing is stricter, as shown in the PR description.

maheshmurag (approved):

skills/claude-api/SKILL.md
felixfbecker: @dev-khalid I think the issue here is that there is a colon inside the string, which then requires quoting to not be ambiguous.

ericharmeling (approved):

Issue #556, opened 2026-03-07
run_eval.py: claude -p never triggers skills/commands (0% trigger rate across all queries)


dthau120391:
## Bug Description

`run_eval.py` creates a command file in `.claude/commands/` with the skill description in YAML frontmatter, then runs `claude -p` with test queries to check if Claude invokes the `Skill` or `Read` tool for that command. In practice, **no query ever triggers the skill** — all should-trigger queries get 0/3 trigger rate, including an explicit "run /lucas:memory-layer please".

## Reproduction

Tested with 8 different skills (144 total queries across all eval sets). Every should-trigger query scored 0/3 triggers. Every should-not-trigger query also scored 0/3 (correctly, but vacuously).

Diagnostic test with maximally obvious trigger:
```json
[{"query": "I need to use the lucas memory layer skill right now, run /lucas:memory-layer please", "should_trigger": true}]
```
Result: 0/3 triggers.

## Environment
- Claude Code v2.1.71
- macOS Darwin 25.3.0
- Model: claude-opus-4-6
- Ran from external terminal (no active Claude Code session)

## Analysis

The issue appears to be that `claude -p` does not auto-trigger commands placed in `.claude/commands/` the same way an interactive Claude Code session triggers skills from `~/.claude/skills/`. Possible causes:

1. **Commands vs Skills**: Commands in `.claude/commands/` may not appear in Claude's `available_skills` list the same way skills do, or may only be invocable via explicit `/command-name` syntax
2. **`claude -p` mode**: The `-p` (pipe/prompt) mode may not load commands/skills at all, or may not present them to the model
3. **Simple query threshold**: Claude may not consult skills for queries it can handle directly — but even explicit "run this skill" failed, ruling this out as the sole cause

## Expected Behavior

At minimum, an explicit request like "run /skill-name please" should trigger the command (rate > 0/3). Description-matched queries should trigger at a reasonable rate.

## Suggested Fix

- Verify that `claude -p` actually loads and presents `.claude/commands/` files to the model
- Consider placing test skills in `~/.claude/skills/` instead of `.claude/commands/` if skills and commands are triggered differently
- Add a `--debug` flag to `run_eval.py` that dumps what Claude actually sees (system prompt, available skills list) to verify the command file is being presented

Curt-Park:
## What I found

I added a command under `.claude/commands` and checked `/skills` in claude code.
It lists the command as a project skill.

## `.claude/commands/test-trigger-abc123.md`
```markdown
---
[...]

PR #996, opened 2026-04-21
fix(skill-creator): close run_eval's 0% recall and harden the loop

ob-aion:
## Summary

Closes the 0% trigger-rate bug in `skill-creator`'s description eval (#556) end to end. `run_eval.py` reported 0% recall for every query, so `run_loop.py` was optimizing against a metric that was always zero.

This builds directly on @MartinCajiao's #1298 — I've folded its isolated-project rewrite in here with credit, added the one fix it was missing, and pinned the whole thing with tests, so there's a single reviewable change instead of scattered halves.

Root causes:

- **Wrong surface.** The candidate was installed as a `.claude/commands/` file, which `claude -p` never auto-triggers. It now installs as a real skill in an isolated, UUID-keyed temp project — the surface that actually auto-triggers.
- **Shadowing (the missing half).** Skill name resolution is **user > project** ([docs](https://code.claude.com/docs/en/skills)), so a same-name skill already under `~/.claude/skills` overrode the candidate and the eval measured the *installed* description. #1298 warns about this; here `run_eval` moves the installed copy aside for the run and restores it after. Without it, recall still reads 0% whenever you optimize a skill you already have installed — the common case.
- **Detection.** Field-scoped, so a short name like `pdf` no longer counts reading `report.pdf` as a hit. Windows stream reading uses reader threads over `select()`, and `shutil.which` resolves the `claude.cmd` shim.
- **Loop robustness.** `run_loop` and `generate_report` crashed on `max()` over empty history for runs that exit before the first iteration; an `improve_description` failure discarded every collected iteration. Both are guarded.

## What's new since #1298

- The user > project shadow is fixed, not just warned.
- Detection is extracted to a pure `_detect_trigger`, and the repo gets its first test scaffolding: 18 offline tests pinning the metric — short-name false positives, Windows paths, explore-then-trigger, shadow move/restore, the loop guards. Stdlib only, no API cost.

Consolidates the scattered attempts at this bug (#1298, #1323). Merging this closes #556 in one pass.

## Test plan

- [ ] `cd skills/skill-creator && python3 -m unittest discover -s scripts/tests` — 18 pass
- [ ] `python -m scripts.run_eval --eval-set <set>.json --skill-path <installed-skill>`: recall reflects the description under test, not the installed copy
- [ ] `python -m scripts.run_loop --eval-set <set>.json --skill-path <skill> --model <m> --max-iterations 0` exits cleanly

Fixes #556. Credit: @MartinCajiao (isolated-project rewrite), @KhalifaGad (field-scoped detection), @Vaikri-costume (loop guards), @xg-gh-25 (review that shaped the isolation).

### DIFF
--- a/skills/skill-creator/scripts/generate_report.py
+++ b/skills/skill-creator/scripts/generate_report.py
@@ -202,8 +202,12 @@ def generate_html(data: dict, auto_refresh: bool = False, skill_name: str = "") 
         <tbody>
 """)
 
-    # Find best iteration for highlighting
-    if test_queries:
+    # Find best iteration for highlighting. A run that exits before the first
+    # history entry (e.g. --max-iterations 0) leaves history empty, and max()
+    # over it would raise.
+    if not history:
+        best_iter = None
+    elif test_queries:
         best_iter = max(history, key=lambda h: h.get("test_passed") or 0).get("iteration")
     else:
         best_iter = max(history, key=lambda h: h.get("train_passed", h.get("passed", 0))).get("iteration")
--- a/skills/skill-creator/scripts/run_eval.py
+++ b/skills/skill-creator/scripts/run_eval.py
@@ -3,26 +3,61 @@
 
 Tests whether a skill's description causes Claude to trigger (read the skill)
 for a set of queries. Outputs results as JSON.
+
+Mechanism: each run creates an isolated temporary project containing the
+candidate description installed as a real skill at
+<tmp>/.claude/skills/<name>/SKILL.md, and `claude -p` executes with that
+project as its working directory. Skills are the only surface `claude -p`
+exposes to the model for auto-triggering (files under .claude/commands/
+appear in slash_commands but never in the model's available_skills list).
+Running in a throwaway project keeps the skill's real name (no suffix) and
+makes concurrent runs collision-free by construction: every run gets its own
+project directory keyed by a full UUID4. The directory is shared by the run's
+parallel workers and removed when the run ends; directories left behind by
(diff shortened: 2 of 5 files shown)


ob-aion:
@maheshmurag @rlancemartin — could one of you take a look when you get a chance? This fixes #556 (run_eval.py reporting a 0% trigger rate) and has been open since April, mergeable with no conflicts.

Update: #1298 independently reworks the same run_eval.py path, and its isolated temp-project approach is the cleaner base — I've followed up there to consolidate. The two are largely complementary: #1298 handles run_eval, while this PR also fixes run_loop.py, where an installed same-name skill shadows the eval and pins recall at 0%. Happy to fold that into #1298 so there's a single thing to review.

ob-aion:
@maheshmurag @rlancemartin — refreshed this into the complete fix for #556 and rebased on current `main`.

It folds in @MartinCajiao's #1298 (the isolated-project `run_eval` rewrite, credited) plus the one thing that PR was missing: skill resolution is **user > project**, so an installed same-name skill shadows the project-level candidate and the eval measures the *installed* description — the reason recall reads 0% whenever you optimize a skill you already have installed. `run_eval` now moves the installed copy aside for the run.

Also adds the loop crash guards (@Vaikri-costume) and the repo's first offline test scaffolding — 18 tests, stdlib only, no API cost. Verified end to end: a matching query now scores 1.0, a non-matching one 0.0.

Merging this closes #556 and supersedes #1298/#1323 in one pass. Happy to adjust anything.

PR #1007, opened 2026-04-22
fix(claude-api): correct compaction context window (200K -> 1M)

gaurav0107:
## Summary

The `claude-api` skill's compaction documentation incorrectly stated that context compaction triggers near the 200K token limit. Compaction is a feature of the 1M token extended context window (available on Sonnet via the `context-1m-2025-08-07` beta header), not the default 200K window.

This fixes the misleading description so users understand when compaction actually triggers.

## Changes

- Updated the compaction section to reference the 1M token context window instead of 200K
- Clarified that compaction requires the extended context beta header

## Test plan

- [x] Verified the fix against Anthropic's public documentation on context compaction
- [x] Confirmed the 1M context window beta header (`context-1m-2025-08-07`) is the correct reference
- [x] No code changes — documentation only

Closes #[ISSUE_NUMBER]

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/skills/claude-api/python/claude-api/README.md
+++ b/skills/claude-api/python/claude-api/README.md
@@ -275,7 +275,7 @@ response2 = conversation.send("What's my name?")  # Claude remembers "Alice"
 
 ### Compaction (long conversations)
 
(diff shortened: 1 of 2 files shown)


gaurav0107:
I do not think its adding any value, closing myself

PR #1145, opened 2026-05-15
Fix model config shape in managed-agents API reference

rlancemartin:
The `model` field on agent create accepts `{id, speed}`, not `{type: "model_config", id, speed}` — there is no `type` discriminator on that object.

Aligns line 41 with the shape already documented at line 206 and in `managed-agents-core.md`.


michael-cohen-io (approved):

ericharmeling (approved):
:shipit:

PR #1183, opened 2026-05-23
fix(xlsx): handle missing AF_UNIX in recalc shim check

### LINKED ISSUE #1120: xlsx recalc.py crashes on Windows: unguarded socket.AF_UNIX access
## Summary

`skills/xlsx/scripts/office/soffice.py` — the `_needs_shim()` function tries to construct an `AF_UNIX` socket to detect whether the LD_PRELOAD shim is needed for sandboxed Linux environments. On Windows Python builds where `socket.AF_UNIX` is not exposed as an attribute, the attribute lookup raises `AttributeError` (not `OSError`), which the surrounding `try/except OSError` does not catch. As a result, `recalc.py` is unusable on these builds — it crashes before any LibreOffice work begins.

## Environment

- Windows 11 Home (Build 26200)
- Python 3.14.4 from a uv-managed install (`cpython-3.14-windows-x86_64-none`)
- LibreOffice 26.2.3.2 at `C:\Program Files\LibreOffice\program\soffice.exe`
- openpyxl 3.1.5
- `anthropics/skills` xlsx skill at current `main`

## Reproduction

1. Build any `.xlsx` with openpyxl that contains formulas.
2. Run `python skills/xlsx/scripts/recalc.py <file.xlsx>`.

**Actual:**

```
  File ".../skills/xlsx/scripts/recalc.py", line 184, in <module>
    main()
  File ".../skills/xlsx/scripts/recalc.py", line 179, in main
    result = recalc(filename, timeout)
  File ".../skills/xlsx/scripts/recalc.py", line 76, in recalc
    if not setup_libreoffice_macro():
  File ".../skills/xlsx/scripts/recalc.py", line 59, in setup_libreoffice_macro
    env=get_soffice_env(),
  File ".../skills/xlsx/scripts/office/soffice.py", line 28, in get_soffice_env
    if _needs_shim():
  File ".../skills/xlsx/scripts/office/soffice.py", line 46, in _needs_shim
    s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
                      ^^^^^^^^^^^^^^
AttributeError: module 'socket' has no attribute 'AF_UNIX'
```

**Expected:** `_needs_shim()` returns `False` on platforms where AF_UNIX is unavailable. The shim itself is Linux-specific (uses `LD_PRELOAD` and a gcc-compiled `.so`), so there is nothing meaningful to shim on Windows. `recalc.py` should then proceed to drive LibreOffice normally.

## Root cause

`_needs_shim()` only catches `OSError`:

```python
def _needs_shim() -> bool:
    try:
        s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
        s.close()
        return False
    except OSError:
        return True
```

On Windows Python builds without AF_UNIX support, `socket.AF_UNIX` is missing as an *attribute* entirely, so the lookup raises `AttributeError` before the constructor runs. AF_UNIX availability on Windows is conditional on both Windows version (10 1803+) and how the interpreter was built — it is not universal even on supported Windows versions.

## Proposed fix

Two-line `hasattr` guard at the top of `_needs_shim()`:

```python
def _needs_shim() -> bool:
    if not hasattr(socket, "AF_UNIX"):
        return False
    try:
        s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
        s.close()
        return False
    except OSError:
        return True
```

Returning `False` early when AF_UNIX is unavailable is correct behavior — `_ensure_shim()` would in any case fail on Windows because it shells out to `gcc` to compile a Linux `.so`, and `LD_PRELOAD` is a Linux-only loader hint.

Optionally also catch `AttributeError` in the existing `except` for defense-in-depth, though the `hasattr` guard is cleaner and more explicit about intent.

## Workaround

Affected users can drive LibreOffice directly, bypassing the skill's macro path:

```
soffice --headless -env:UserInstallation=file:///<isolated-profile> --calc --convert-to xlsx --outdir <tmpdir> <file.xlsx>
```

Confirmed working on the affected environment. The `--convert-to xlsx` round-trip forces a recalculate-and-resave, achieving the same outcome as the skill's macro-driven recalc.

superShen0916:
## Summary
- Return false from the LibreOffice shim check when the Python socket module does not expose AF_UNIX
- Add a focused regression test for Windows builds where AF_UNIX is unavailable

Fixes #1120

## Test plan
- python3 skills/xlsx/scripts/office/test_soffice.py
- python3 -m py_compile skills/xlsx/scripts/office/soffice.py skills/xlsx/scripts/office/test_soffice.py

Note: I also tried unittest discovery under skills/xlsx/scripts/office, but it imports validators that require defusedxml, which is not installed in this local environment.


superShen0916:
Updated the branch to current `main` without rewriting history. The focused regression test and `py_compile` both pass, and the PR diff remains limited to `soffice.py` plus its test.

Issue #1167, opened 2026-05-19
pptx skill produces .pptx files that fail to import into Apple Keynote (silent OOXML compliance issues)


MYCL94:
### Summary

The `pptx` skill produces `.pptx` files that **cannot be imported into Apple Keynote** on macOS. The same files open without complaint in Microsoft PowerPoint, PowerPoint Online, and LibreOffice Impress, so the failure is silent until a Mac user actually tries to open the file.

The user-visible error in Keynote is:

> "Das Dateiformat ist ungültig" / "The file format is invalid"
> or
> "kann nicht importiert werden" / "cannot be imported"

with no further diagnostic. Quick Look (spacebar preview) also fails to render the file.

After a clean OneDrive round-trip (upload → download → open in Keynote), the file imports correctly, confirming the issue is in the file content the skill produces — PowerPoint Online sanitizes it silently.

### Affected scope

Any `.pptx` produced by `python-pptx` via the `pptx` skill that is intended to be opened on macOS in Keynote. Estimated impact: a significant fraction of macOS users in Cowork mode hit this with no recourse other than installing PowerPoint.

### Root cause (three issues)

Diffing a `python-pptx` output against the same file after a PowerPoint Online round-trip identifies three structural OOXML problems:

#### 1. Contradictory `<p:sldSz>` element on widescreen decks

`python-pptx` ships with a 4:3 default template. When user code does:

```python
prs.slide_width = Inches(13.333)
prs.slide_height = Inches(7.5)
```

the `cx` attribute on `ppt/presentation.xml`'s `<p:sldSz>` is updated, but the `type` attribute is **not**. The resulting element is:

```xml
<p:sldSz cx="12191695" cy="6858000" type="screen4x3"/>
```

— widescreen dimensions explicitly declared as `screen4x3`. Keynote's import validator compares the dimensions against the `type` and rejects the contradiction. PowerPoint Online rewrites the element to:

```xml
<p:sldSz cx="12192000" cy="6858000"/>
```

(no `type` attribute, snapped cx to the canonical 16:9 widescreen value of 12,192,000 EMU).

#### 2. Missing `<p:notesMasterIdLst>` in `presentation.xml`

[...]

PR #1231, opened 2026-06-01
fix(pptx): add Keynote/macOS OOXML compliance sanitizer

Oxygen56:
## Summary

Adds an explicit PPTX compatibility command for decks written directly by python-pptx or pptxgenjs. The documented file-check flow runs it before validation.

The command normalizes slide-size metadata, registers a missing notes master in schema order, and removes Windows printer settings while preserving embedded OLE parts and their content types. It writes through an atomic replacement and leaves already-sanitized files unchanged.

Fixes #1167.

## Verification

- Reproduced the original notes-master ordering failure against the bundled PresentationML schema.
- Reproduced the shared `.bin` content-type regression on a deck containing an embedded workbook.
- Direct python-pptx save → sanitizer CLI → repository validator passed for a plain notes deck, an embedded-workbook deck, and an equivalent shared `.bin` declaration.
- Speaker notes and embedded workbook bytes are preserved; a second sanitizer run does not change the resulting file bytes.
- All 13 focused regressions passed, including interrupted-write preservation.
- Actual Keynote import was not retested because Keynote is not installed on this host.

This version is based on current main, where `pack.py` has been removed; it uses the explicit command for all delivery flows.


oi8io:
Thanks for picking this up — the diagnosis in #1167 is solid and this is worth landing. I ran the patch against the reproduction case from the issue and hit three problems. Two are correctness bugs that I think should block merge; the third is a coverage gap.

## 1. `_add_notes_master_id` inserts the element in the wrong position

`<p:notesMasterIdLst>` is appended after `</p:sldIdLst>`, but ECMA-376 `CT_Presentation` declares an `xsd:sequence`:

```
sldMasterIdLst → notesMasterIdLst → handoutMasterIdLst → sldIdLst → sldSz → notesSz → …
```

It has to come *before* `sldIdLst`. Validating the patched file against the bundled `schemas/ISO-IEC29500-4_2016/pml.xsd`:

```
$ python -c "
from lxml import etree; import zipfile
s = etree.XMLSchema(etree.parse('skills/pptx/scripts/office/schemas/ISO-IEC29500-4_2016/pml.xsd'))
x = etree.fromstring(zipfile.ZipFile('test.pptx').read('ppt/presentation.xml'))
print(s.validate(x)); print(s.error_log)
"
False
Element '{...presentationml/2006/main}notesMasterIdLst': This element is not
expected. Expected is one of ( {...}sldIdLst, {...}sldSz, ... )
```

Element order actually matters here — `skills/pptx/SKILL.md` already warns about it:

> **Never reorder the children of `<p:presentation>`.** … move the element and the same deck becomes unopenable.

So this trades a Keynote failure for a potential PowerPoint failure. Keynote appears lenient about the ordering, which is probably why it wasn't caught.

Worth noting: `scripts/office/validate.py` reports `All validations PASSED!` on the malformed output, so this would ship silently.

### Suggested diff

```python
 def _add_notes_master_id(pres_xml: str, rels_xml: str) -> str:
     if "<p:notesMasterIdLst" in pres_xml:
         return pres_xml
     m = re.search(r'<Relationship Id="(rId\d+)"[^>]*?notesMaster[^>]*?/>', rels_xml)
     if not m:
         return pres_xml
     rid = m.group(1)
     block = f'<p:notesMasterIdLst><p:notesMasterId r:id="{rid}"/></p:notesMasterIdLst>'
-    return re.sub(r"</p:sldIdLst>", "</p:sldIdLst>" + block, pres_xml, count=1)
+    # ECMA-376 CT_Presentation is an xsd:sequence — notesMasterIdLst must
+    # precede sldIdLst. Insert before the first successor that is present.
+    for name in ("handoutMasterIdLst", "sldIdLst", "sldSz", "notesSz"):
+        anchor = re.search(rf"<p:{name}\b", pres_xml)
+        if anchor:
+            return pres_xml[:anchor.start()] + block + pres_xml[anchor.start():]
+    anchor = re.search(r"</p:sldMasterIdLst>", pres_xml)
+    if anchor:
+        return pres_xml[:anchor.end()] + block + pres_xml[anchor.end():]
+    return pres_xml
```

## 2. `_strip_printer_settings` corrupts decks containing OLE objects

```python
entries[ct_path] = re.sub(r'<Default Extension="bin"[^>]*/>', "", ct).encode("utf-8")
```

The `bin` default is removed unconditionally, but embedded OLE objects are also `.bin` (`ppt/embeddings/oleObject1.bin`) and depend on that declaration. On a deck with an embedded Excel sheet or equation object, the part survives while its content type disappears — which is exactly the class of breakage this PR is trying to fix.

Reproduction: a package containing both `ppt/printerSettings/printerSettings1.bin` and `ppt/embeddings/oleObject1.bin`, after `sanitize()`:

```
oleObject1.bin present : True
<Default Extension="bin"> present : False   ← orphaned part
```

### Suggested diff

```python
     ct_path = "[Content_Types].xml"
-    if ct_path in entries:
+    # Only drop the `bin` default when nothing else needs it — OLE objects
+    # (ppt/embeddings/oleObject1.bin) rely on this declaration.
+    if ct_path in entries and not any(
+        n.lower().endswith(".bin") for n in entries
+    ):
         ct = entries[ct_path].decode("utf-8")
         entries[ct_path] = re.sub(
             r'<Default Extension="bin"[^>]*/>', "", ct
         ).encode("utf-8")
```

## 3. The hook in `pack.py` doesn't cover the reported failure path

`pack()` runs on the unzip → edit → rezip flow. The reproduction in #1167 — and in this PR's own test plan — is `python-pptx` calling `prs.save()` directly, which never goes through `pack.py`. The "before" and "after" snippets in the test plan operate on two different files, so the repro case isn't actually shown to be fixed.

Suggestions, in rough order of preference:

- Expose it as a CLI (`python scripts/keynote_sanitize.py deck.pptx`) and have `SKILL.md` require it in the QA section next to `validate.py`, so it covers decks built with python-pptx or pptxgenjs too.
- Or fold the check into `validate.py` so a non-compliant deck is at least reported.
- Keeping the `pack.py` hook as well is fine — it just isn't sufficient on its own.

## Two smaller things

**Non-atomic write.** `sanitize()` reopens the same path with `ZipFile(path, "w")`, which truncates immediately. An exception between truncate and the final `writestr` leaves the user with a destroyed deck. Writing to a `NamedTemporaryFile` in the same directory and `shutil.move`-ing over the original makes it crash-safe.

**No tests.** Given both bugs above are one assertion away from being caught, some coverage seems worthwhile. I have a 13-case suite (schema-order assertion, OLE regression, idempotency, 4:3 passthrough, custom sizes, crash safety) that I am happy to hand over or open as a PR against your branch — whichever you prefer. Just say the word.

---

Verified against `python-pptx` 1.0.2 and the bundled ISO-IEC29500-4:2016 schemas.

98zc5g5jyw-arch:
Hi @Oxygen56 — thanks for the rework. Reviewed at head `eb75d1a01b635a4c08fd02da83b76d395ea5c868` and re-ran the previous review's three cases plus a fresh end-to-end pass locally (uv, Python 3.12.13, python-pptx 1.0.2, lxml against the bundled ISO-IEC29500-4:2016 schemas).

✅ **Ordering (previous finding 1).** On the #1167 repro — python-pptx deck at 13.333 in, i.e. `<p:sldSz cx="12191695" cy="6858000" type="screen4x3"/>`, a notesMaster relationship with no registration, printer settings present — the repaired `ppt/presentation.xml` now validates against `scripts/office/schemas/ISO-IEC29500-4_2016/pml.xsd`. The counterfactual (appending the block after `</p:sldIdLst>`) fails with exactly the error quoted in the last review, and `test_notes_master_is_inserted_before_slide_list` pins the sequence order.

✅ **Shared `.bin` / OLE (previous finding 2).** On a deck carrying `ppt/embeddings/oleObject1.bin` (Override already present) and `oleObject2.bin` (reachable only from a slide relationship), both parts survive byte-identical, both end up with an `oleObject` Override, and the printer-settings `Default` is dropped only when nothing else needs it. An unaccounted `.bin` (`ppt/media/mystery.bin`) fails closed: exit 1, the offending part is named in the message, and the original file is byte-identical.

✅ **Coverage (previous finding 3).** The repair is now a CLI plus a SKILL.md QA step, which covers the direct python-pptx/pptxgenjs save path the earlier pack-time hook missed.

✅ **Atomic write and idempotency.** The simulated-interrupted-write test passes; a second run prints "No Keynote compatibility changes needed" and the file hash is unchanged. 13/13 tests pass (`python -m unittest discover -s tests`, and under pytest).

✅ **Change scope.** On the repro exactly one part is removed (`ppt/printerSettings/printerSettings1.bin`) and exactly three are rewritten (`ppt/presentation.xml`, `ppt/_rels/presentation.xml.rels`, `[Content_Types].xml`); every other part — notes included — is byte-identical, entry order is preserved, `python-pptx` reopens the result with notes and slide size intact, and `scripts/office/validate.py` passes before and after. A clean default deck only loses its printer settings; its canonical 4:3 `sldSz` is left untouched.

Non-blocking suggestions:

1. A `notesMasterIdLst` that exists but sits out of sequence is not repaired: `_add_notes_master_id` returns early, so the file stays schema-invalid (confirmed — still fails `pml.xsd` with the element-order error) while the CLI still prints "Sanitized" for the other repairs. That is precisely the state the earlier revision of this PR left behind. You already have the successor list, so moving the element when it appears after the first successor — or at minimum reporting it and exiting non-zero — would close that hole.
2. `_patch_slide_size` strips `type` for every non-canonical size, including consistent ones: an A4 deck (`cx="10692000" cy="7560000" type="A4"`) comes back as `<p:sldSz cx="10692000" cy="7560000"/>` with the dimensions preserved. Rendering-neutral, but slightly wider than "only what Keynote rejects" — limiting the removal to cases where `type` contradicts the geometry would keep the write smaller.
3. `validate.py` still prints "All validations PASSED!" on both the pre-sanitize repro and the misordered deck, so nothing in the flow turns red if the sanitizer is skipped. A `<p:presentation>` child-order check in the pptx validator would make that gap non-silent.
4. `tests/` has no `__init__.py`, so `python -m unittest discover -s tests -t .` raises `ImportError: Start directory is not importable` (`discover -s tests`, `python tests/test_keynote_sanitize.py` and pytest all work) — trivial, but cheap to make every invocation work.

One caveat I can't close: Keynote is not available on this host, so the macOS-side claim remains unverified here; everything above is OOXML-side, plus the schema and consumer checks. Nice work on the fail-closed paths and the CRC-checked atomic write.

98zc5g5jyw-arch:
Re-verified at head `37034b6` — both follow-up suggestions are implemented and pinned by tests:

- ✅ Out-of-sequence `notesMasterIdLst` (suggestion 1): detected via the namespace-aware child list and refused — the new test drives the real CLI and asserts exit 1, the diagnostic on stderr, no "Sanitized" claim, and a byte-identical file. That is the "report and exit non-zero" minimum the note asked for.
- ✅ Consistent non-`screen4x3` sizes (suggestion 2): `type` is now removed only when a screen-aspect type contradicts the geometry; the A4 deck round-trips unchanged (`test_consistent_a4_type_is_preserved`).
- ✅ `tests/__init__.py` (suggestion 4): every invocation now works — `discover -s tests`, `discover -s tests -t .`, and pytest.
- ✅ Full suite re-run on my side under two environments (Python 3.11 + defusedxml + lxml via uv, and a local 3.11): 15/15 pass under both unittest and pytest.
- The validator-gap note (suggestion 3) was informational; the SKILL.md text now makes the fail-closed expectation explicit, which covers the practical path.

LGTM from my side; the earlier caveat stands (no Keynote on this host, so everything above is OOXML-side).

Issue #1221, opened 2026-05-30
skill-creator: description-optimization (run_loop) is broken on Windows - claude not found + select() on pipes


crichter163-droid:
## Summary
The skill-creator's description-optimization tooling (`scripts/run_loop.py` →
`scripts/run_eval.py` + `scripts/improve_description.py`) cannot run on Windows.
Every `claude -p` subprocess fails with `FileNotFoundError: [WinError 2]`, and
the streaming reader uses `select.select()` on a subprocess pipe, which is
POSIX-only. On Windows all queries error out and the loop crashes when it
reaches the "Improving description" step.

## Environment
- Windows 11, Python 3.14
- Claude Code CLI 2.1.158 (installed via npm; resolves to `claude.cmd`)
- skills repo at commit `da20c92`

## Reproduction
```
python -m scripts.run_loop \
  --eval-set <trigger-eval.json> \
  --skill-path <some-skill> \
  --model <model-id> --max-iterations 1 --verbose
```
Result: `Warning: query failed: [WinError 2] Das System kann die angegebene Datei nicht finden`
repeated for every query, then a traceback ending in
`FileNotFoundError: [WinError 2]` from `improve_description.py`.

## Root causes & fixes

**1. `subprocess` can't launch the CLI by bare name `"claude"` on Windows.**
On Windows the CLI is `claude.cmd`; `CreateProcess` cannot execute a `.cmd`
directly and bare `"claude"` isn't resolved. Fix: resolve via
`shutil.which("claude")` and, on `os.name == "nt"`, route through the command
interpreter (`%COMSPEC% /c ...`). Applies to both `run_eval.py` (line ~71) and
`improve_description.py` (line ~26).

**2. `select.select()` on a pipe is POSIX-only (`run_eval.py` line ~108).**
On Windows `select` only works with sockets, so the streaming read loop can
never read the subprocess stdout. Fix: drop `select`; iterate
`for raw_line in process.stdout:` (portable) and enforce the timeout with a
`threading.Timer` watchdog that kills the process. Early-trigger detection logic
is unchanged.

## Patch
A unified diff against `da20c92` is attached (`skill-creator-windows-fixes.patch`),
touching only `scripts/run_eval.py` and `scripts/improve_description.py`. After
applying, `run_loop` launches `claude -p` successfully and streams events on
Windows.

## Note (separate, lower priority)
With the subprocess fixed, the trigger-detection harness still reported
`recall=0%` on Windows: the temp slash-command written to
[...]

PR #1232, opened 2026-06-01
fix(skill-creator): add Windows subprocess compatibility for run_loop

Oxygen56:
## Summary

On Windows, the skill-creator's description-optimization tooling (`scripts/run_loop.py` → `run_eval.py` + `improve_description.py`) fails with two errors:

1. **`claude` not found** — the CLI resolves to `claude.cmd` on Windows, which `CreateProcess` cannot launch directly by bare name
2. **`select.select()` on pipes is POSIX-only** — the streaming read loop uses `select()` which only works with sockets on Windows

## Fix

Two small changes (29 insertions, 27 deletions):

| File | Change |
|---|---|
| `improve_description.py` | Resolve `claude` via `shutil.which()`; route through `%COMSPEC% /c` on `os.name == "nt"` |
| `run_eval.py` | Same claude resolution + COMSPEC routing; replace `select()` loop with portable `process.stdout` line iteration + `threading.Timer` watchdog |

The fix is portable — on macOS/Linux, `shutil.which("claude")` returns the same path as before and `os.name == "nt"` is false, so the code path is unchanged.

## Test plan

On Windows with Claude Code CLI installed via npm:
```
python -m scripts.run_loop --eval-set eval.json --skill-path ../my-skill --model claude-sonnet-4-6 --max-iterations 1 --verbose
```
Before: `FileNotFoundError: [WinError 2]` on every query
After: queries evaluate successfully ✓

On macOS/Linux: no behavior change (verified by inspection — the `os.name == "nt"` branch never executes).

Closes #1221.

### DIFF
--- a/skills/skill-creator/scripts/improve_description.py
+++ b/skills/skill-creator/scripts/improve_description.py
@@ -10,6 +10,7 @@ import argparse
 import json
 import os
 import re
+import shutil
 import subprocess
 import sys
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


Oxygen56:
Hi maintainers, these 5 PRs have been open for 15-16 days with no reviews or CI checks triggered. Would appreciate if someone could take a look:

- #1226 fix: eval_metadata.json path mismatch
- #1227 fix: xlsx Windows AF_UNIX compatibility
- #1230 fix: encoding=utf-8 for Windows write_text/open
- #1231 fix: Keynote/macOS OOXML compliance
- #1232 fix: Windows subprocess run_loop

Thanks!

Issue #1260, opened 2026-06-04
skill-creator: trigger eval writes {skill}-skill-{hash} command files into the live project's .claude/commands/, polluting the skill registry of concurrent sessions


alvingarcia:
## Summary

`skill-creator`'s trigger evaluation (`skills/skill-creator/scripts/run_eval.py`) tests
description variants by writing synthetic command files named `{skill_name}-skill-{8hex}.md`
directly into the **user's live project** `.claude/commands/` directory. With the default
`--num-workers 10`, up to ~10 of these files exist simultaneously during the eval window.

Every concurrent Claude Code session rooted at that project sees them: they appear in the
available-skills system reminders, the `Skill` tool's listing, and the interactive `/` typeahead
— 10+ near-identical hash-suffixed duplicates per skill under eval, rotating as workers cycle.

Each file is deleted in a `finally` after its query completes, so the pollution is transient —
which makes it confusing rather than obviously broken: users see phantom skills that vanish when
they look for them on disk.

## Root cause (verified at `main`, 2026-06-04)

In `skills/skill-creator/scripts/run_eval.py`:

- `find_project_root()` walks **up from cwd to the nearest real `.claude/`** — i.e. the user's
  live project (lines ~22–32).
- `run_single_query()` writes the variant command file to
  `project_root/.claude/commands/{skill}-skill-{uuid8}.md` (lines ~51–54, 68) and launches
  `claude -p` with `cwd=project_root` (lines ~85–91).
- `run_eval()` fans out via `ProcessPoolExecutor(max_workers=num_workers)` (line ~198), default
  10 (line ~264) — so the live `commands/` dir holds up to ~10 synthetic entries at once.
- Cleanup is `command_file.unlink()` per query in `finally` (lines ~179–181): cleaned **after**
  each query, never **isolated during** it.

`run_loop.py` inherits the behavior — it imports `find_project_root`/`run_eval` from
`run_eval.py`, so both eval entry points are affected.

## Impact

1. **Registry pollution for concurrent sessions** — any other Claude Code session (human or
   agent) rooted at the project sees the synthetic variants in its skill list during the eval
   window and may invoke or recommend the wrong entry.
2. **Typeahead noise** — the interactive `/` palette shows N hash-suffixed duplicates of the
   skill under eval.
3. **Context waste** — each variant description (~900 chars) is injected into every
   system-reminder skill listing of every live session, multiplied by worker count.
[...]

PR #1261, opened 2026-06-04
fix(skill-creator): isolate trigger-eval command files from the live project registry

alvingarcia:
## Summary

Fixes #1260 — the trigger eval writes its synthetic `{skill}-skill-{8hex}.md` command files
into the **user's live project** `.claude/commands/` (via `find_project_root()` + `cwd=project_root`),
so during the parallel eval window (default 10 workers) every concurrent Claude Code session
rooted at that project sees up to ~10 phantom hash-suffixed skill variants in its registry,
typeahead, and system-reminder skill listings.

This change isolates each eval variant in a per-query throwaway project root:

- variant command file is written under `tempfile.mkdtemp()/.claude/commands/`
- the `claude -p` subprocess runs with `cwd=eval_root` (it discovers commands from its cwd's
  `.claude/`; global auth/config in `~/.claude` are unaffected by cwd)
- `finally` removes the whole temp root (`shutil.rmtree`, replacing the per-file `unlink`)

The eval becomes hermetic: no live-registry pollution at any worker count, and project-level
commands/skills in the user's repo can no longer confound the trigger measurement. (This also
resolves the project-level case of anthropics/claude-plugins-official#632; skills installed
globally in `~/.claude/skills/` are still loaded regardless of cwd and would need a HOME
override — out of scope here.)

`run_loop.py` inherits the fix — it delegates to `run_eval()`. `find_project_root()` and the
`project_root` parameter are now unused for command placement; left intact to keep this diff
minimal, happy to remove them here or in a follow-up if preferred.

## Test plan

- [x] `python3 -m py_compile scripts/run_eval.py`
- [x] Ran the patched eval from a directory inside a live Claude Code project (where the old
  code reproducibly wrote variants into the project's `.claude/commands/`):
  1-query eval set, `--num-workers 1 --runs-per-query 1` → **1/1 PASS** (variant discovered and
  triggered from the isolated root), the live project's `.claude/commands/` was never created,
  and no `skill-eval-*` temp dirs were left behind.

### DIFF
--- a/skills/skill-creator/scripts/run_eval.py
+++ b/skills/skill-creator/scripts/run_eval.py
@@ -56,7 +56,9 @@ def run_single_query(
     """
     unique_id = uuid.uuid4().hex[:8]
     clean_name = f"{skill_name}-skill-{unique_id}"
-    eval_root = Path(tempfile.mkdtemp(prefix=f"skill-eval-{unique_id}-"))
+    # Include the skill name in the prefix so any tempdir stranded by a
+    # killed worker (rmtree never runs on SIGKILL) is attributable.
+    eval_root = Path(tempfile.mkdtemp(prefix=f"skill-eval-{skill_name}-{unique_id}-"))
     command_file = eval_root / ".claude" / "commands" / f"{clean_name}.md"
 
     try:
--- a/skills/skill-creator/scripts/run_eval.py
+++ b/skills/skill-creator/scripts/run_eval.py
@@ -9,8 +9,10 @@ import argparse
 import json
 import os
 import select
+import shutil
 import subprocess
 import sys
+import tempfile
 import time
 import uuid
 from concurrent.futures import ProcessPoolExecutor, as_completed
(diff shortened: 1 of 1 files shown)


alvingarcia:
Friendly ping on this one — it's been open ~3.5 weeks with no maintainer review. It's a small (+15/−8, single file), self-contained fix for the registry-pollution bug in #1260, with a green smoke test (variant discovered/triggered from the isolated root, live project `.claude/commands/` never written, no temp dirs leaked). A community reviewer on #1260 also independently validated the per-query tempdir approach as sound.

Happy to fold in two optional refinements from that #1260 discussion if you'd like them in this PR:

1. Prefix the eval root for forensics, so a crashed run is attributable:
   ```python
   eval_root = Path(tempfile.mkdtemp(prefix=f"skilleval-{skill_name}-{unique_id}-"))
   ```
2. Log the isolation so the hermetic-by-design intent is explicit in eval logs:
   ```python
   logger.info(f"Isolated eval root: {eval_root} (cleaned on exit)")
   ```

I can also remove the now-unused `find_project_root()` / `project_root` parameter in this PR or a follow-up — whichever you prefer to keep the diff minimal. Just let me know the direction and I'll push.

alvingarcia:
Update: I've folded in both refinements from the #1260 discussion so the diff is final and ready to review.

1. **Forensic tempdir prefix** — the eval root is now `skill-eval-{skill_name}-{unique_id}-*`, so a tempdir stranded by a SIGKILLed worker is attributable to a specific skill/run.
2. **Isolation log line** (this push, `492a1f8`) — a module-level `logger.info("Isolated eval root: %s (cleaned on exit)", eval_root)` makes the hermetic-by-design intent explicit in eval logs. It uses `logging.getLogger(__name__)` with no handler configured, so it's silent by default and only surfaces when the caller enables logging — no new stderr noise across the parallel workers.

Still a small, single-file change (`run_eval.py`), green smoke test unchanged (variant discovered/triggered from the isolated root; live `.claude/commands/` never written; no temp dirs leaked).

I deliberately left the dead-code cleanup of `find_project_root()` / the `project_root` parameter out of this PR to keep the diff minimal, since `run_loop.py` also imports `find_project_root` and threads `project_root` through — removing it cleanly would touch that file too. Happy to do it as a small follow-up if you'd like the pass-through gone. Just let me know.

PR #1297, opened 2026-06-09
Update claude-api skill: scheduled deployments, vault env-var credentials, system.message events

rlancemartin:
## What this does

Adds the latest Managed Agents documentation to the claude-api skill. Follow-up to #1294, which covered the Claude Fable 5 model updates; this PR carries the Managed Agents half of the same documentation refresh.

## What changed

- **Scheduled deployments**: new `shared/managed-agents-scheduled-deployments.md` covering cron schedules and DST semantics, deployment runs, pause/auto-pause, and lifecycle; Deployments and Deployment Runs endpoints added to the API reference; beta-header coverage and SKILL.md reading-guide routing updated.
- **Vault `environment_variable` credentials**: a new credential type for non-MCP services (CLIs, SDKs, direct API calls) — the sandbox sees an opaque placeholder and the real secret is substituted at egress, gated by per-credential networking allowlists. Secrets guidance in the tools, client-patterns, and onboarding docs rewritten around this; self-hosted sandbox caveats noted (not yet supported there).
- **`system.message` events**: update the agent's system prompt mid-session (Claude Opus 4.8 only), documented in the events guide and API reference with constraints and validation errors.

## Verification

Per-file diff verified against the upstream documentation source; the only intentional divergences are the SKILL.md frontmatter, the `purpose`-field removals from Files API examples (#1081), and LICENSE.txt.

### DIFF
--- a/skills/claude-api/SKILL.md
+++ b/skills/claude-api/SKILL.md
@@ -249,7 +249,7 @@ For placement patterns, architectural guidance, and the silent-invalidator audit
 
 **Mandatory flow:** Agent (once) → Session (every run). `model`/`system`/`tools` live on the agent, never the session. See `shared/managed-agents-overview.md` for the full reading guide, beta headers, and pitfalls.
 
-**Beta headers:** `managed-agents-2026-04-01` — the SDK sets this automatically for all `client.beta.{agents,environments,sessions,vaults,memory_stores}.*` calls. Skills API uses `skills-2025-10-02` and Files API uses `files-api-2025-04-14`, but you don't need to explicitly pass those in for endpoints other than `/v1/skills` and `/v1/files`.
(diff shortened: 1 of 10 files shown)


rlancemartin:
Cross-checked this against claude-cli-internal main file-by-file: ✅ all CMA content is present and byte-identical (modulo the expected `{{OPUS_ID}}` → literal substitutions), including the post-launch corrections (scheduled-deployments failure behavior, `error.type` discriminator, `system.message` idle-with-stop_reason wording). Two small items:

1. The new SKILL.md routing line says deployments come "with run records, **retries, and auto-pause**" — those two claims were corrected away upstream (no documented retry-on-transient; archived/deleted agent **auto-archives** the deployment, no auto-pause). This was stale on internal main too — just fixed there (claude-cli-internal#41314, commit 00029ea5a4). Suggest: "with per-firing run records and lifecycle controls (pause/unpause/archive)".
2. Pre-existing drift, not from this PR: `client-patterns.md`'s upload example reads `files.upload({ file })` but internal has `files.upload({ file, purpose: 'agent_resource' })` (from claude-cli-internal#26587). Worth picking up while this file is open.

ericharmeling (approved):

PR #1298, opened 2026-06-10
fix(skill-creator): isolate trigger evals and handle Windows and runtime failures

### LINKED ISSUE #556: run_eval.py: claude -p never triggers skills/commands (0% trigger rate across all queries)

MartinCajiao:
## Problem

Trigger evaluation can report false misses or invalid scores: per-worker command probes compete, select() on subprocess pipes fails on Windows, and unrelated tools stop the scan. Runtime failures also become non-triggers, incorrectly passing negative examples and misleading optimization.

Fixes #556.

## What changes

- Install one real candidate skill at `<temporary project>/.claude/skills/<name>/SKILL.md`, preserving its production name and candidate description. Parallel workers share it; concurrent runs use distinct UUID-based roots.
- Leave caller files untouched. The legacy `project_root` argument remains accepted but is no longer used.
- Reject same-name personal skill collisions, including those under `CLAUDE_CONFIG_DIR`.
- Resolve the Claude executable, including Windows npm shims. Read stdout/stderr with portable threads instead of select(). Close stdin and cap probes at eight turns.
- Keep scanning after Bash/Glob and other orientation tools. Match exact Skill names or skill-entrypoint Read paths, including Windows separators.
- Stop the probe process tree during cleanup: `taskkill /T /F` on Windows; a separate session/process group on POSIX.
- Treat error results, timeouts, truncated output and worker failures as invalid evaluations. Abort the batch instead of assigning zero trigger rates. Cancel pending work where possible; active workers finish their bounded cleanup.
- Preserve successful evaluation JSON. Initial optimization failures propagate; later failures preserve completed iterations. Empty and partial reports render correctly.
- Write candidate artifacts as UTF-8.

## Validation

Latest revision: Windows, Python 3.14.

```sh
# From skills/skill-creator
python -m unittest discover -s scripts/tests -v
uvx ruff check scripts/run_eval.py scripts/tests/test_eval_failures.py scripts/tests/test_stream_probe.py
git diff --check
```

- **21/21 tests pass.** Coverage includes artifact isolation, exact matching, late triggers, configuration collisions, invalid batch rejection, cleanup and partial-result recovery.
- A real Windows .cmd wrapper launches a Python child; the regression verifies that the child cannot continue after timeout. This test requires process-termination permissions and was run outside the restricted sandbox.
- Ruff passes for the files changed in the latest revision. Whitespace checks pass.
- Tests make no model calls and need no credentials. Instructions and limitations are in `scripts/tests/README.md`.

## Scope and limitations

The temporary project excludes caller-project files and configuration. Queries requiring local fixtures need those fixtures made available separately. Personal and managed Claude configuration is not fully isolated.

The original investigation included live routing experiments on Claude Code 2.1.149. Those historical observations are not a claim about every current version. This revision has not rerun live model evaluation or POSIX integration. The candidate uses the standard skill layout independently of legacy command behavior.

Related work: #794, #996, #1027, #1755 and #1757. This PR combines a real-skill artifact with portable stream handling, full-turn detection and explicit failure semantics.

### DIFF
--- a/skills/skill-creator/scripts/generate_report.py
+++ b/skills/skill-creator/scripts/generate_report.py
@@ -203,7 +203,9 @@ def generate_html(data: dict, auto_refresh: bool = False, skill_name: str = "") 
 """)
 
     # Find best iteration for highlighting
-    if test_queries:
+    if not history:
+        best_iter = None
+    elif test_queries:
         best_iter = max(history, key=lambda h: h.get("test_passed") or 0).get("iteration")
     else:
         best_iter = max(history, key=lambda h: h.get("train_passed", h.get("passed", 0))).get("iteration")
--- a/skills/skill-creator/scripts/generate_report.py
+++ b/skills/skill-creator/scripts/generate_report.py
@@ -219,7 +219,7 @@ def generate_html(data: dict, auto_refresh: bool = False, skill_name: str = "") 
         test_total = h.get("test_total")
         description = h.get("description", "")
         train_results = h.get("train_results", h.get("results", []))
-        test_results = h.get("test_results", [])
+        test_results = h.get("test_results") or []
 
         # Create lookups for results by query
         train_by_query = {r["query"]: r for r in train_results}
--- a/skills/skill-creator/scripts/run_eval.py
+++ b/skills/skill-creator/scripts/run_eval.py
@@ -34,6 +34,10 @@ from scripts.utils import parse_skill_md
 # runs where Claude explores (TodoWrite/Glob/Bash) before invoking the skill.
 MAX_CLAUDE_TURNS = 8
 
+# Leftover artifacts must be at least this old before the sweep removes
+# them, so it never deletes the live artifact of a concurrent eval run.
+STALE_ARTIFACT_MIN_AGE_SECONDS = 3600
+
 _EVAL_SUFFIX_RE = re.compile(r"-eval-[0-9a-f]{8}$")
 
 
@@ -159,7 +163,12 @@ def run_single_query(
             except Empty:
                 continue
             if raw is None:  # EOF: process finished and output is drained
-                process.wait()
+                try:
+                    process.wait(timeout=10)
+                except subprocess.TimeoutExpired:
+                    # stdout closed but the process lingers; the finally
+                    # block kills it
+                    return False
                 if process.returncode != 0:
                     print(
                         f"Warning: claude -p exited with code {process.returncode} "
@@ -227,9 +236,14 @@ def _cleanup_stale_artifacts(skills_dir: Path, skill_name: str) -> None:
     """Remove eval skills left behind by a previous crashed/killed run."""
     if not skills_dir.is_dir():
         return
+    cutoff = time.time() - STALE_ARTIFACT_MIN_AGE_SECONDS
     for entry in skills_dir.glob(f"{skill_name}-eval-*"):
         if entry.is_dir() and _EVAL_SUFFIX_RE.search(entry.name):
-            shutil.rmtree(entry, ignore_errors=True)
+            try:
+                if entry.stat().st_mtime < cutoff:
+                    shutil.rmtree(entry, ignore_errors=True)
+            except OSError:
+                pass
 
 
 def _warn_if_shadowed(skill_name: str, project_root: Path) -> None:
--- a/skills/skill-creator/scripts/run_eval.py
+++ b/skills/skill-creator/scripts/run_eval.py
@@ -4,21 +4,27 @@
 Tests whether a skill's description causes Claude to trigger (read the skill)
 for a set of queries. Outputs results as JSON.
 
-Mechanism: the candidate description is installed once per run as a real
-skill at <project_root>/.claude/skills/<name>-eval-<id>/SKILL.md, because
-that is the only surface `claude -p` exposes to the model for
-auto-triggering (files under .claude/commands/ appear in slash_commands but
-never in the model's available_skills list, so they can never auto-trigger).
-The artifact is shared by all parallel workers and removed when the run ends.
+Mechanism: each run creates an isolated temporary project containing the
+candidate description installed as a real skill at
+<tmp>/.claude/skills/<name>/SKILL.md, and `claude -p` executes with that
+project as its working directory. Skills are the only surface `claude -p`
+exposes to the model for auto-triggering (files under .claude/commands/
+appear in slash_commands but never in the model's available_skills list).
+Running in a throwaway project keeps the skill's real name (no suffix),
+never touches the user's project, and makes concurrent runs collision-free
+by construction: every run gets its own project directory keyed by a full
+UUID4. The directory is shared by the run's parallel workers and removed
+when the run ends; directories left behind by killed runs are swept once
+they are older than --stale-artifact-hours.
 """
 
 import argparse
 import json
 import os
-import re
 import shutil
 import subprocess
 import sys
+import tempfile
 import threading
 import time
 import uuid
@@ -34,11 +40,13 @@ from scripts.utils import parse_skill_md
 # runs where Claude explores (TodoWrite/Glob/Bash) before invoking the skill.
 MAX_CLAUDE_TURNS = 8
 
-# Leftover artifacts must be at least this old before the sweep removes
-# them, so it never deletes the live artifact of a concurrent eval run.
-STALE_ARTIFACT_MIN_AGE_SECONDS = 3600
+# All eval projects live under this root in the OS temp directory.
+EVAL_PROJECTS_ROOT = Path(tempfile.gettempdir()) / "claude-skill-evals"
 
-_EVAL_SUFFIX_RE = re.compile(r"-eval-[0-9a-f]{8}$")
+# Leftover eval projects must be at least this old (hours) before the sweep
+# removes them: long enough to survive a paused debugging session, short
+# enough not to accumulate cruft. Override with --stale-artifact-hours.
+DEFAULT_STALE_ARTIFACT_HOURS = 12.0
 
 
 def find_project_root() -> Path:
@@ -90,13 +98,14 @@ def run_single_query(
     query: str,
     eval_skill_name: str,
     timeout: int,
-    project_root: str,
+    eval_project_dir: str,
     model: str | None = None,
     claude_cli: str | None = None,
 ) -> bool:
     """Run a single query and return whether the eval skill was triggered.
 
-    Streams `claude -p` output and returns True as soon as a Skill/Read
+    Runs `claude -p` with the isolated eval project as its working
+    directory, streams its output, and returns True as soon as a Skill/Read
     tool_use referencing the eval skill appears. Other tool calls
     (TodoWrite, Glob, Bash, ...) are ignored rather than treated as
     non-triggers, since Claude often explores before consulting a skill.
@@ -123,7 +132,7 @@ def run_single_query(
         cmd,
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
-        cwd=project_root,
+        cwd=eval_project_dir,
         env=env,
     )
 
@@ -232,13 +241,18 @@ def run_single_query(
             process.wait()
 
 
-def _cleanup_stale_artifacts(skills_dir: Path, skill_name: str) -> None:
-    """Remove eval skills left behind by a previous crashed/killed run."""
-    if not skills_dir.is_dir():
+def _sweep_stale_eval_projects(stale_hours: float) -> None:
+    """Remove eval projects left behind by previous crashed/killed runs.
+
+    Only directories older than the threshold are removed, so the sweep
+    never deletes the live project of a concurrent eval run and survives
+    paused debugging sessions.
+    """
+    if not EVAL_PROJECTS_ROOT.is_dir():
         return
-    cutoff = time.time() - STALE_ARTIFACT_MIN_AGE_SECONDS
-    for entry in skills_dir.glob(f"{skill_name}-eval-*"):
-        if entry.is_dir() and _EVAL_SUFFIX_RE.search(entry.name):
+    cutoff = time.time() - stale_hours * 3600
+    for entry in EVAL_PROJECTS_ROOT.iterdir():
+        if entry.is_dir():
             try:
                 if entry.stat().st_mtime < cutoff:
                     shutil.rmtree(entry, ignore_errors=True)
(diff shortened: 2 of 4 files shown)


xg-gh-25:
Brutal diagnosis — four independent causes each sufficient for 0% recall, which explains why partial fixes didn't move the number. The "optimizing against noise" framing is exactly right: a metric that always reads zero might as well not exist.

The cross-platform testing rigor (Windows stream handling, executable resolution) and the systematic verification strategy (stub harness + offline tests before real integration) are both production patterns we recognize. A few observations:

**On the artifact installation strategy:** The decision to install as `.claude/skills/<name>-eval-<id>/SKILL.md` instead of swapping the real skill in-place is the safer choice. The tradeoff you note (suffix on the skill name) is acceptable because routing happens on *description content*, not the name. One risk to flag: if the model's routing logic includes **edit distance on skill names** as a tie-breaker (some frameworks do this), the suffix could deflate trigger rates slightly. Have you verified that Claude Code's skill picker ignores name similarity entirely?

**On cause 3 (exploration counted as miss):** The fix — "non-Skill tools are ignored, only terminal events decide not-triggered" — is correct for auto-triggering, but it changes what you're measuring. Before: "did the skill trigger *first*?" After: "did the skill trigger *eventually*?" The latter is the right metric for description optimization (you want the model to reach the skill at all), but if you're also evaluating routing *efficiency* (how many irrelevant tools were used before the right one), you'd want to track that separately. Consider logging `steps_to_trigger` as optional metadata.

**On cause 4 (parallel worker race):** The shared artifact solution is clean, but there's a hidden complexity: if multiple *runs* execute concurrently (e.g., CI pipeline with parallel jobs), they'll still collide on the artifact name. The `-eval-<id>` suffix mitigates this if `<id>` is globally unique (UUID), but if it's process-scoped (PID), you can get cross-process races. Suggest documenting the `<id>` uniqueness guarantee — or add a lockfile if not already present.

**On the verification strategy:** The offline stub harness is excellent (we use this pattern too — "canned stream-json" for zero-cost iteration). One gap: your offline tests cover Windows stream handling and parallel workers, but they don't cover the actual *trigger detection logic* against malformed/truncated `tool_use` events. Consider adding a fuzzer that injects partial JSON chunks mid-event to verify your detection doesn't false-positive.

Minor: The PR description says "Tradeoff kept deliberately: the eval skill's name carries an -eval-<id> suffix while production skills don't." But you also say "the suffix avoids collisions with installed skills." These are compatible, but clarify: does the suffix prevent shadowing (good), or does it introduce name-mismatch risk (potential concern)? The warning about shadowing suggests you're aware of both sides.

Last: The "sweep stale `<name>-eval-*` artifacts" logic is defensive (we do this too). What's the age threshold? 2 hours? 24 hours? If it's too aggressive, you'll delete artifacts from paused debugging sessions. If too permissive, you'll accumulate cruft. Consider making it configurable or at least documenting the default.

---
*Eval loop signal integrity recognized via [SwarmAI](https://github.com/xg-gh-25/SwarmAI). Discussion: [T-MEM](https://github.com/xg-gh-25/SwarmAI/discussions/13)*

MartinCajiao:
Hi Xiaogang, thanks for the rigorous review! Great catches here. You are
exactly right: these are architectural trade-offs and polishing points that
I am happy to adjust. Here is the breakdown and how I plan to address them
right away:

*1. On the artifact installation strategy (Suffix vs. Name-Mismatch):* The
suffix was designed primarily to *prevent shadowing* and ensure isolation
between production skills and evaluation instances. Regarding the
name-mismatch risk: Claude Code's skill picker routes tools based on the
description metadata rather than strict edit distance on names, so the
-eval- suffix doesn't degrade trigger rates in standard testing. However,
to eliminate any edge-case risk of name-mismatch or shadowing completely, I
am going to refine this in the next commit so the evaluation context
strictly shadows the name only within a sandboxed temporary environment,
keeping the production namespace completely clean.

*2. On the cleanup logic (Configurable Threshold):* You’re spot on. Right
now, the age threshold for sweeping stale artifacts is hardcoded to a
defensive 2-hour window. To avoid killing active debugging sessions while
preventing disk cruft, *I am going to turn this into an adjustable
parameter*.

   -

   I will expose it as a configurable option (e.g., --eval-clean-threshold
   or via CLAUDE.md / config context).
   -

   I will default it to 12 hours to balance safety for long debugging
   sessions and automatic cleanup, and document it clearly in the README/docs.

*3. On the parallel worker race condition (Cause 4):* To ensure absolute
safety in concurrent environments (like heavily parallelized CI pipelines),
I will upgrade the process-scoped ID resolution to a globally unique
identifier (UUID v4) or implement a lightweight lockfile mechanism (.lock)
within the local evaluation directory to completely eliminate cross-process
collisions.

I’ll push a commit addressing these refinements in a bit. Thanks again for
the feedback!

MartinCajiao:
Pushed 79b3df9 delivering the refinements discussed above:

**1. Sandboxed eval environment (name-mismatch + shadowing):** went one step further than promised. Each run now creates an isolated throwaway project under the OS temp dir (`<tmp>/claude-skill-evals/<skill>-<uuid4>/.claude/skills/<real-name>/SKILL.md`) and `claude -p` executes with that project as its working directory. The skill keeps its **real name** -- isolation comes from the project directory, not from a name suffix -- so the edit-distance question is moot by construction, and the caller's project is never touched by eval artifacts at all. On your original question: empirically the suffixed artifact was already triggering 2/2 on matching queries in the Windows e2e (routing follows the description), but removing the suffix removes the question entirely.

**2. Configurable sweep threshold:** now `--stale-artifact-hours`, defaulting to 12h as proposed. One correction to my earlier reply: the previous hardcoded window was 1 hour, not 2. The sweep also only ever operates inside the dedicated temp root now, so it cannot touch anything outside it.

**3. Cross-run collisions:** to be precise, the previous id was already UUID-derived (a uuid4 hex prefix), not process-scoped -- but the new layout makes the guarantee structural rather than probabilistic: a full 32-hex UUID4 keys a dedicated per-run project directory, so concurrent runs (parallel CI jobs included) cannot collide regardless of id entropy. No lockfile needed.

**On `steps_to_trigger`:** agreed -- that is the right metric for routing *efficiency* as opposed to routing *reach*. Kept out of this PR to keep the output schema stable for `run_loop.py`/`improve_description.py`, but it is a clean additive follow-up.

**On fuzzing the detection against malformed/truncated `tool_use` events:** the streaming detector accumulates `input_json_delta` fragments into a buffer and substring-matches on the accumulated state, so split/partial chunks are handled by construction, and malformed JSON lines are skipped. I would like to land that as in-repo tests, but the scripts currently have no test scaffolding -- verification lives in an external stub harness (canned stream-json, zero API cost) for now. Happy to contribute test scaffolding as a separate PR if maintainers want it.

Re-verified end to end on Windows 11 + Claude Code 2.1.149 after the change: 2/2 should-trigger, 0/2 should-not-trigger, zero leftover directories in temp.

ob-aion:
Nice work here — the isolated temp-project approach is cleaner than what I did in #996, and the Windows fixes (threads+queue over select(), shutil.which for the .cmd shim) plus the --max-turns cap are clear wins. Glad to see this be the run_eval base.

One gap worth flagging: because this keeps the candidate's real skill name in the temp project, a same-named skill already installed under ~/.claude/skills still shadows the eval — personal skills load regardless of the eval's working directory, so claude -p sees two skills with that name and recall can still read as 0%. That's the other half of what #996 fixes, in run_loop.py: it moves an installed same-name skill aside for the duration of the run.

So the two are complementary — your run_eval rework plus that run_loop guard. Happy to fold the guard into this PR (I can open a PR against your branch, co-authored) or send it as a follow-up once this lands, whichever you'd rather. Just want the 0% gone.

KhalifaGad:
Thanks for the comprehensive rework — installing the candidate as a real skill plus the shadow warning addresses root causes that the single-file patches (including my now-closed #1340) don't touch.

One small thing on the matcher itself. `_tool_use_mentions` does:

```python
return eval_skill_name in json.dumps(tool_input)
```

That substring-matches the name against the *entire* serialized input, including `args`/`file_path`. For a short, common skill name it over-counts — a skill named `pdf` registers a "trigger" on any `Skill`/`Read` call whose input merely mentions a `.pdf`:

```python
>>> 'pdf' in json.dumps({'command': 'convert report.pdf'})
True
```

Field-scoping avoids it without losing any real detection — exact-match the `skill` field for `Skill`, and match `Read` on a path component (since you install under `.claude/skills/<name>/`):

```python
def _tool_use_mentions(eval_skill_name, tool_name, tool_input):
    if tool_name == "Skill":
        return (tool_input.get("skill") or "").strip() == eval_skill_name
    if tool_name == "Read":
        return f"/{eval_skill_name}/" in (tool_input.get("file_path") or "")
    return False
```

Same idea for the streamed path: parsing the completed `accumulated_json` and field-checking it beats substring-matching the raw bytes (which include argument text).

Verified on Claude Code 2.1.140: with field-scoped exact-match, real triggers still detect (8/10 on a 20-query adversarial set, the other 2 being genuine model self-serve), and the short-name false-positive class drops out. Happy to send this as a small follow-up against your branch if it's useful.

MartinCajiao:
Thanks @KhalifaGad — great catch. The substring match was clearly wrong for short names.

Fixed in ce1adf8. I took your field-scoped approach and extended it to cover the whole detection flow:

- **Skill** → `tool_input["skill"]` must equal the skill name exactly
- **Read** → `/<name>/` must appear in `file_path`

Two things on top of the snippet:

1. **The substring check also lived in the streaming path** (`content_block_delta` / `content_block_stop`), which is the path that actually decides most runs — the `assistant` event is only a fallback. Fixing just `_tool_use_mentions()` would have left the false positive alive there. So I moved the decision to `content_block_stop`, where the full input JSON is available, and routed it through the same `_tool_use_mentions()` helper. The delta handler now only accumulates, since partial JSON can't be field-scoped.

2. **Windows path separators.** Before the `Read` check I normalize `\` to `/`, because on Windows `claude -p` emits backslash paths like `...\.claude\skills\<name>\SKILL.md`, where `"/<name>/"` would never match — which would have quietly reintroduced 0% recall on the exact platform this PR is fixing.

Your adversarial 20-query set is a nice idea. Happy to fold a short-name false-positive case into the offline stub harness as a follow-up if you think it's worth pinning down in tests.

Vaikri-costume:
Thank you for this.  This is a very thorough fix for `run_eval.py` and has been super helpful. Two companion bugs in the files this PR doesn't touch that are worth folding in, since the cause 3 + 4 fixes make them more reachable:

**1. `run_loop.py` + `generate_report.py` — empty-history `max()` crash**

`run_loop()` picks the best iteration with `max(history, …)`. On the degenerate path
(`--max-iterations 0`, or `improve_description` crashing before the first
`history.append`), `history` is empty and `max()` raises
`ValueError: max() arg is an empty sequence`. `generate_report.generate_html` has the
same unguarded `max()`, so guarding only `run_loop` moves the crash one frame up.
Your multi-turn exploration fix makes the early-exit path more reachable — any run that
exits before the first history append now hits this.

Repro:
```
python -m scripts.run_loop --eval-set <set>.json --skill-path <skill>   --model <m> --max-iterations 0
# → ValueError: max() arg is an empty sequence
```

Fix for `run_loop.py` — insert before `# Find the best iteration` (~L216):
```python
    if not history:
        return {
            "exit_reason": exit_reason,
            "original_description": original_description,
            "best_description": original_description,
            "best_score": None,
            "best_train_score": None,
            "best_test_score": None,
            "final_description": current_description,
            "iterations_run": 0,
            "holdout": holdout,
            "train_size": len(train_set),
            "test_size": len(test_set),
            "history": history,
        }
```

Fix for `generate_report.py` — guard the highlight `max()` (~L205):
```python
    if not history:
        best_iter = None
    elif test_queries:
        best_iter = max(history, key=lambda h: h.get("test_passed") or 0).get("iteration")
    else:
        best_iter = max(history, key=lambda h: h.get("train_passed", h.get("passed", 0))).get("iteration")
```

**2. `run_loop.py` — `improve_description` crash discards all accumulated history**

If `improve_description()` raises mid-run (rate limit, model error, credential expiry),
the exception propagates uncaught and loses every iteration collected so far. Wrapping in
try/except and breaking gracefully saves partial results.

Fix — wrap the `improve_description()` call:
```python
        try:
            new_description = improve_description(...)
            improve_elapsed = time.time() - t0
        except Exception as e:
            exit_reason = f"improve_description crashed on iteration {iteration}: {e}"
            if verbose:
                print(f"\nError during description improvement — saving partial results.\n  {e}", file=sys.stderr)
            break
```

Feel free to fold either or both in — no credit needed.

KhalifaGad:
Awesome — thanks for taking it the rest of the way. You're right that the streaming path is the real decider; my snippet only covered the `assistant` fallback, so moving the field-scoped decision to `content_block_stop` is the correct call. And the Windows backslash normalization is a great catch — that would've silently reintroduced 0% recall on the exact platform this PR targets.

Yes, the short-name false-positive is worth pinning in the stub harness — happy to share the set; a skill named `pdf` triggering on `convert report.pdf` is the cleanest repro. Thanks for the thoughtful response! 🙏

98zc5g5jyw-arch:
Flagging a merge-collision risk before this grows further: this PR edits `skills/skill-creator/scripts/run_eval.py` and `run_loop.py`, which overlap with two other open PRs — #1755 (eval-harness isolation redesign; run_eval.py + run_loop.py + SKILL.md) and #1757 (full-turn trigger-scan fix; run_eval.py). #1735 covered the same ground and was closed unmerged. Three open PRs editing the same files will conflict at merge time; recommend maintainers pick an order or ask the authors to coordinate. The added failure-path tests are valuable regardless of which design lands.

MartinCajiao:
> Flagging a merge-collision risk before this grows further: this PR edits `skills/skill-creator/scripts/run_eval.py` and `run_loop.py`, which overlap with two other open PRs — #1755 (eval-harness isolation redesign; run_eval.py + run_loop.py + SKILL.md) and #1757 (full-turn trigger-scan fix; run_eval.py). #1735 covered the same ground and was closed unmerged. Three open PRs editing the same files will conflict at merge time; recommend maintainers pick an order or ask the authors to coordinate. The added failure-path tests are valuable regardless of which design lands.

Thanks for flagging the overlap. I compared #1298 directly against #1755 and #1757 and synced this branch with current main.

There is one architectural distinction that I think is important when choosing the integration base:

#1755 and #1757 still build the evaluation probe under `.claude/commands/`, while #1298 installs the candidate as an actual `.claude/skills/<name>/SKILL.md` skill.

That distinction is the original root cause behind the 0%-recall failure. In the `claude -p` init event, command files appear under `slash_commands`, while the model-facing auto-trigger registry is populated from real skills. Improving full-turn scanning around a command probe fixes detector behavior, but it can still leave the harness measuring a different surface from the one being optimized.

I added regression coverage to #1298 that explicitly pins this invariant: the eval artifact must be a real skill, keep the production name/description, never fall back to `.claude/commands`, remain isolated from the caller’s project, and use collision-free temporary roots.

I also folded in two useful ideas from the later PRs without changing that architecture: the headless subprocess now uses `stdin=DEVNULL`, and there is explicit regression coverage for non-skill tool activity (e.g. Bash/Glob) occurring before a later Skill/Read trigger.

The later PRs still contain useful ideas worth incorporating:

* #1755: fixture ergonomics and init-event collision diagnostics.
* #1757: focused TriggerDetector-style regression testing.

But #1298 additionally covers the Windows `select()` pipe failure, npm `claude.cmd` resolution, portable stream reading, exact field-scoped Skill/Read matching (including Windows paths), subprocess error visibility, parallel-worker correctness, and `run_loop.py` / `generate_report.py` failure recovery.

Given that, I think #1298 is the strongest integration base: it fixes the artifact being measured as well as the detector and cross-platform/runtime failure modes. The smaller ergonomics/testing improvements from #1755/#1757 can then be folded onto it without regressing to a slash-command probe.

98zc5g5jyw-arch:
Thanks for the detailed comparison — that rationale is useful, and I independently re-verified the current head `d53172aa` on macOS (python3, 3.11):

- ✅ **New suite passes**: from `skills/skill-creator`, `python3 -m unittest discover -s scripts/tests -v` → **21 tests, all pass** (1 skipped = the Windows-only `taskkill` regression). Stdlib only, no Claude credentials — matches `scripts/tests/README.md`.
- ✅ **The claimed invariants are pinned by tests**: artifact isolation (`test_artifact_is_never_written_into_callers_project`, `test_parallel_runs_get_distinct_project_roots`), real-skill artifact (`test_candidate_is_installed_as_real_skill`, `test_eval_skill_keeps_real_name_and_description`), closed stdin (`test_orientation_tool_then_skill_still_triggers_and_stdin_is_closed`), invalid measurements rejected instead of scored as 0% (`test_error_result_is_not_a_negative_measurement`, `test_truncated_stream_is_an_error_even_with_exit_zero`), and run_loop failure-path preservation (`test_eval_failure_preserves_completed_iteration`, `test_initial_eval_failure_still_raises`).
- ✅ **Branch state**: `mergeable: MERGEABLE` after your sync with main.
- Scope note (same as the tests README): this does not establish routing accuracy for a particular Claude Code version — that still needs a live positive/negative evaluation with the version recorded.

On merge order: the overlap is unchanged — this PR edits `run_eval.py` + `run_loop.py` (+ report generation), #1755 edits `run_eval.py` + `run_loop.py` + `SKILL.md`, and #1757 edits `run_eval.py`; all three are still open, so a deliberate pick-an-order remains worth doing before any of them lands. The command-probe vs real-skill-artifact distinction you describe is concrete and now test-pinned, which is the strongest thing a reviewer can ask for here; the sequencing call itself is the maintainers'.

PR #1323, opened 2026-06-16
fix(skill-creator): run_eval trigger detection misses real skill name and bails on first non-Skill tool

Polluelo978:
## Problem

`scripts/run_eval.py::run_single_query` fails to detect that a skill triggered, causing the description-optimization loop (`run_loop.py`) to report **recall=0%** for every should-trigger query. With all candidates tied at 0, the loop just returns the original description and never actually optimizes.

Two root causes in the stream detection:

1. **Only the registered command name is matched.** The function registers a temp command named `{skill}-skill-{uuid}` and only checks whether that random name (`clean_name`) appears in the tool input. But an installed skill fires under its **real name** (e.g. `cve-intel`), so the random name never matches -> always False.
2. **Bails to False on the first non-Skill/Read tool.** Real Claude Code runs often call `Bash`/`WebFetch`/etc. before consulting the skill. The `content_block_start` handler does `return False` on the first such tool, missing a later Skill/Read call.

## Fix

- Match the **real `skill_name`** in addition to the registered `clean_name`, in both the streaming path and the full-assistant-message fallback.
- Do **not** return False on the first non-Skill/Read tool. Reset the pending state and keep scanning until `type == "result"`.

Early-return on a positive match keeps positives fast; negatives run to `result` (short Q&A), bounded by `--timeout`.

## Evidence (same skill + eval set, 3 runs/query, `--model claude-sonnet-4-6 --timeout 120`)

| metric    | before | after (held-out test) |
|-----------|--------|-----------------------|
| recall    | 0%     | 100% |
| precision | 100%   | 100% |
| accuracy  | 50%    | 100% |

After the fix the loop can measure description quality as intended (held-out test 8/8).

## Notes

- No behavioral change for skills not installed locally: `clean_name` matching is preserved; `skill_name` matching is additive.
- Single-file change in `skills/skill-creator/scripts/run_eval.py`.

### DIFF
--- a/skills/skill-creator/scripts/run_eval.py
+++ b/skills/skill-creator/scripts/run_eval.py
@@ -138,20 +138,27 @@ def run_single_query(
                                     pending_tool_name = tool_name
                                     accumulated_json = ""
                                 else:
-                                    return False
+                                    # Don't bail: other tools (Bash/WebFetch/etc.)
+                                    # may precede a later Skill/Read call. Reset the
+                                    # pending state and keep scanning.
+                                    pending_tool_name = None
 
                         elif se_type == "content_block_delta" and pending_tool_name:
                             delta = se.get("delta", {})
                             if delta.get("type") == "input_json_delta":
                                 accumulated_json += delta.get("partial_json", "")
-                                if clean_name in accumulated_json:
+                                # Also match the real skill name: an installed skill
+                                # fires under its actual name, not the temp command
+                                # name (clean_name).
+                                if clean_name in accumulated_json or skill_name in accumulated_json:
                                     return True
 
                         elif se_type in ("content_block_stop", "message_stop"):
                             if pending_tool_name:
-                                return clean_name in accumulated_json
-                            if se_type == "message_stop":
-                                return False
+                                if clean_name in accumulated_json or skill_name in accumulated_json:
+                                    return True
+                                pending_tool_name = None
+                            # Don't return False here; keep scanning until "result".
 
                     # Fallback: full assistant message
                     elif event.get("type") == "assistant":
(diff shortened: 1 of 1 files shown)


Vaikri-costume:
Strong +1 — this fixes a bug that fully breaks the description optimizer in a common setup.

With a user-global `~/.claude/CLAUDE.md` that calls `AskUserQuestion` (or any clarifying tool) before acting, the model's *first* action is never `Skill`/`Read`, so the current `else: return False` scores every query — including genuine triggers — as not-triggered. Measured effect: `run_loop.py` reports **0% recall for every description**, so the optimizer ends up "improving" descriptions against an all-zeros signal.

After this PR's scan-past-non-skill-tools logic, the measurements become meaningful again — two skills I tested split to a true **17% vs 100%** recall instead of a uniform 0% (the 17% one is a skill whose task the model self-serves; the 100% one needs a specialized methodology, so it consults reliably). Same intent as the earlier #1208.

Vaikri-costume:
Update: PR #1298 (MartinCajiao) addresses the same detection issues plus the architectural cause (artifact install path), Windows stream reading, and parallel worker race. It field-scopes the trigger match (vs. substring here) and is more complete overall. Worth reviewing #1298 as the potential merge target if maintainers are looking at both.

skills/skill-creator/scripts/run_eval.py
```diff
@@ -138,20 +138,27 @@ def run_single_query(
                                     pending_tool_name = tool_name
                                     accumulated_json = ""
                                 else:
-                                    return False
+                                    # Don't bail: other tools (Bash/WebFetch/etc.)
+                                    # may precede a later Skill/Read call. Reset the
+                                    # pending state and keep scanning.
+                                    pending_tool_name = None
```
alpha819beta, line 144: No return value here?

PR #1331, opened 2026-06-20
fix: Windows cp1252 encoding crashes in DOCX validator (non-ASCII print, binary XML read) 

danabases:
Bug 1 — docx.py line 249: The → character can't be encoded by Windows console (cp1252). Fix: use ASCII ->
Bug 2 — base.py line 763: open(xml_file, "r") with no encoding specified defaults to cp1252 on Windows. numbering.xml contains byte 0x8f (undefined in cp1252), so it crashes when reading. Fix: open in binary mode so lxml handles XML encoding detection itself.


danabases:
do over

Issue #1338, opened 2026-06-21
skill-creator eval: trigger detection never matches (randomized clean_name vs. invoked skill name) → 0% trigger rate on Claude Code 2.1.x


KhalifaGad:
## Summary

`skill-creator`'s trigger evaluation (`scripts/run_eval.py`) reports a **0% trigger rate for every query** — including obvious, on-the-nose matches — on Claude Code `2.1.140`. Because every query (positive and negative) is scored "not triggered," `run_eval`/`run_loop` and the **description optimizer** are effectively non-functional: every candidate description scores identically (100% precision / 0% recall), so the optimizer has no signal to improve on.

The skill **does** trigger correctly; the detector just never recognizes it.

## Environment

- Claude Code CLI: `2.1.140`
- `skill-creator`: current `main`
- Model: `claude-opus-4-8`
- Python: 3.13

## Root cause

`run_single_query()` installs the skill under a randomized command name and stores it as `clean_name`:

```python
clean_name = f"{skill_name}-skill-{unique_id}"   # e.g. "change-delivery-skill-38eea5"
```

Detection then checks whether **`clean_name`** appears in the skill invocation — both in the streamed deltas and in the full assistant message:

```python
# stream path
if clean_name in accumulated_json:
    return True
# assistant-message path
if tool_name == "Skill" and clean_name in tool_input.get("skill", ""):
    triggered = True
```

But Claude Code invokes the skill by its **normalized base name**, not the randomized filename. Captured `tool_use` from a real run (a textbook should-trigger query):

```json
{"skill": "change-delivery", "args": "Feature: Add webhook retries ... phased plan with subagent execution and reviewer approval gates ..."}
```

So the matcher is comparing `"change-delivery-skill-38eea5"` against `"change-delivery"`. The randomized `-skill-<uuid>` suffix means `clean_name` is **never** a substring of the invoked name — the check is `False` 100% of the time. (It's also directionally inverted: the invoked name is a substring of `clean_name`, not the other way around.)

## Proof it's the matcher (not the model or the description)

Same model, same 20-query eval set, single run each; the **only** change is the matcher:

| Matcher | should-trigger detected | should-not-trigger |
|---|---|---|
| stock (`clean_name in …`) | **0 / 10** | 10 / 10 (pass by default — nothing ever triggers) |
| match base skill name instead | **8 / 10** | 10 / 10 |

The negatives "pass" in the stock version only because *nothing* is ever detected as triggered.

## Proposed fix

[...]

PR #1339, opened 2026-06-21
fix(skill-creator): match real skill name in eval trigger detection (#1338)

KhalifaGad:
Fixes the trigger-detection bug reported in #1338.

## Problem

`run_eval.py`'s `run_single_query()` installs the skill under a randomized command name stored in `clean_name`:

```python
clean_name = f"{skill_name}-skill-{unique_id}"   # e.g. "change-delivery-skill-38eea5"
```

Detection then checks whether **`clean_name`** appears in the skill invocation. But Claude Code invokes the skill by its **real name**, not the randomized filename. Captured `tool_use` from a real run:

```json
{"skill": "change-delivery", "args": "..."}
```

`"change-delivery-skill-38eea5"` is never a substring of `"change-delivery"`, so the match is `False` 100% of the time. Every query — including textbook matches — scores "not triggered," making `run_eval`, `run_loop`, and the description optimizer non-functional on current Claude Code: every candidate description scores identically (100% precision / 0% recall), so the optimizer has no signal.

## Fix

Add a field-scoped `_skill_triggered()` helper, used by **both** detection paths (streamed events and full assistant message):

- **Skill** calls: exact-match the real skill name (`skill_name`) — or `clean_name` — on the `skill` field.
- **Read** calls: match on `file_path`.

Field-scoping (rather than substring-matching the whole serialized input) avoids the *inverse* failure mode: a short, common skill name like `test` or `pdf` would otherwise match echoed argument text and over-count triggers. The streamed early-detection is preserved, but the stream path can now only early-return `True` — the full-message path stays authoritative, so it can't reintroduce the original false-negative.

## Verification

Claude Code `2.1.140`, model `claude-opus-4-8`, the only change being this patch:

| Matcher | should-trigger | should-not-trigger |
| --- | --- | --- |
| before (`clean_name` only) | **0 / 10** | 10 / 10 (pass only because nothing ever triggers) |
| after (this PR) | **8 / 10** | 10 / 10 |

The remaining 2 should-triggers are genuine model behavior (it self-serves those tasks), not detection misses. Additionally:
- a topically-adjacent negative (`"rename our deliveryPlan variable to plan…"`) stays correctly untriggered, and
- a unit check confirms a `pdf`-named skill no longer false-positives on a `convert report.pdf` argument.

## Note for maintainers

Verified on Claude Code `2.1.140`, where the model invokes via `{"skill": "<skill_name>"}`. The helper exact-matches that field (plus `clean_name`); if other CLI versions use a different invocation shape, only `_skill_triggered()` needs adjusting.

Refs #1338


KhalifaGad:
Reopening wasn't possible — this PR's branch history got corrupted by an accidental force-push (the diff exploded to the whole tree). I've rebuilt the branch with correct history and opened a clean replacement: https://github.com/anthropics/skills/pull/1340. Apologies for the noise.

PR #1386, opened 2026-07-03
Fix mcp-builder evaluation harness: handle parallel tool calls and missing text responses

ardalann:
## Summary
Fixes two bugs in `skills/mcp-builder/scripts/evaluation.py`.

## Bug 1: Parallel tool call handling
When Claude returns multiple `tool_use` blocks in one turn, the harness only executed the first (`next(...)`) and sent a single `tool_result`. The API requires a `tool_result` for every `tool_use`, so the next request failed with a 400.

**Fix:** Collect all `tool_use` blocks, execute them concurrently with `asyncio.gather`, and return all `tool_result` blocks in one user message. This is safe over a single connection: `mcp.ClientSession` multiplexes concurrent requests by request ID.

## Bug 2: Missing text response crash
When the final assistant message has no text block, `response_text` was `None`. `extract_xml_content(None, ...)` then raised `TypeError`.

**Fix:** Join all text blocks via `response_text_from_content()` (empty string fallback) and guard `extract_xml_content` against `None`/empty input. Joining all text blocks (instead of taking only the first) also means `<response>` tags split across multiple text blocks are found, and `agent_loop`'s declared `tuple[str, ...]` return type is now accurate.

## Testing
- `python3 -m py_compile skills/mcp-builder/scripts/evaluation.py`
- The repo has no test harness for these scripts; verified by syntax check and review against the Messages API tool-use contract.

### DIFF
--- a/skills/mcp-builder/scripts/evaluation.py
+++ b/skills/mcp-builder/scripts/evaluation.py
@@ -100,9 +100,10 @@ async def execute_tool_use(
     connection: Any,
     tool_metrics: dict[str, Any],
 ) -> dict[str, Any]:
-    """Execute one tool call and return the MCP tool_result block."""
+    """Execute one tool call and return the tool_result content block."""
     tool_name = tool_use.name
     tool_input = tool_use.input
+
     tool_start_ts = time.time()
     try:
         tool_result = await connection.call_tool(tool_name, tool_input)
@@ -111,10 +112,12 @@ async def execute_tool_use(
         tool_response = f"Error executing tool {tool_name}: {str(e)}\n"
         tool_response += traceback.format_exc()
     tool_duration = time.time() - tool_start_ts
+
     if tool_name not in tool_metrics:
         tool_metrics[tool_name] = {"count": 0, "durations": []}
     tool_metrics[tool_name]["count"] += 1
     tool_metrics[tool_name]["durations"].append(tool_duration)
+
     return {
         "type": "tool_result",
         "tool_use_id": tool_use.id,
--- a/skills/mcp-builder/scripts/evaluation.py
+++ b/skills/mcp-builder/scripts/evaluation.py
@@ -76,11 +76,50 @@ def parse_evaluation_file(file_path: Path) -> list[dict[str, Any]]:
         return []
 
 
-def extract_xml_content(text: str, tag: str) -> str | None:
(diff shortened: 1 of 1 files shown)


WillO-OlliW:
Hit both of these independently and can confirm the diagnosis and the fixes.

**Bug 1** reproduces immediately with any current model. `claude-sonnet-5` issues parallel tool calls, so the run died on task 1:

```
anthropic.BadRequestError: 400 - messages.2: `tool_use` ids were found without
`tool_result` blocks immediately after: toolu_01WpWVsZ3GQ6koHsLrYD2uq2
```

The same patch you describe — collect every `tool_use` block, return all `tool_result` blocks in one user message — fixed it. Worth noting the harness still defaults to `claude-3-7-sonnet-20250219`, which is presumably why this went unnoticed: the default model rarely triggers it, so anyone passing `-m` with something current is the first to hit it.

**Bug 2** I hit as the follow-on you would expect, and there is a root cause underneath it that this PR does not address.

The no-text turn was not random — it was `max_tokens=4096` (lines 99 and 140). Once tool calls actually return data (i.e. after #1390 is fixed), responses grow, the model's `<summary>`/`<feedback>`/`<response>` block is long by construction of `EVALUATION_PROMPT`, and the turn gets truncated. Raising it to 16384 made the condition disappear entirely for me.

That matters for this PR because **the empty-string fallback converts a crash into a silent wrong answer**. After your fix, a truncated turn yields `response_text = ""`, `extract_xml_content` returns `None`, the report prints `N/A`, and `score` is 0 — indistinguishable from the model genuinely failing the task. The crash was at least honest about being a harness problem.

Suggestion: keep the fallback, but say what happened rather than scoring it. I used

```python
if response_text is None:
    response_text = f"<response>NO_TEXT_BLOCK stop_reason={response.stop_reason}</response>"
```

so the report carries `stop_reason=max_tokens` in the answer column instead of a plausible-looking failure. Raising `max_tokens` as well would stop it arising in the first place.

Same underlying theme as #1390: the harness's failures are consistently rendered as low scores rather than as errors, which is how an invalid run ends up looking like a measurement.

Environment: Windows 11, Python 3.13.14, `mcp` 1.29.1, `anthropic` 1.0.0, model `claude-sonnet-5`, stdio server exposing 66 tools. With all of #1390, this PR, `max_tokens`, and the report-write bug (#1669) fixed, the suite ran clean end to end at 10/10.

Issue #1075, opened 2026-04-30
possible XSS in skill-creator's templates.


cruzcunha:
I've used a security extension for Gemini CLI on the skill-creator, and the extension reported vulnerabilities on scripts that embeds a JSON on a '<script>' tags without escaping '<' and '>'. 
I know this is probably a minor issue, but i'm open one just to let you know. 

 ## Vulnerability: Cross-Site Scripting (XSS)                                                                                                                                                              │
Line Content: return template.replace("/*__EMBEDDED_DATA__*/", f"const EMBEDDED_DATA = {data_json};")                                                                                                  Description: The `generate_review.py` script embeds a JSON string into a `<script>` tag inside the HTML template without escaping `<` and `>`. If the JSON contains a string with `</script><script>alert(1)</script>`, it can break out of the script tag and execute arbitrary Javascript in the user's browser when they view the review page.                                        │

Recommendation: Escape the JSON string before embedding it into the HTML document. Specifically, `data_json = json.dumps(embedded).replace("<", "\\u003c").replace(">", "\\u003e").replace("&", "\\u0026")` should be used.

PR #1393, opened 2026-07-05
fix(skill-creator): escape HTML in eval-viewer embedded JSON

griffithsbs:
Fixes #1075.

### Problem

`eval-viewer/generate_review.py` builds the standalone results viewer by
embedding the eval data into `viewer.html` inside a `<script>` block:

```python
data_json = json.dumps(embedded)
return template.replace("/*__EMBEDDED_DATA__*/", f"const EMBEDDED_DATA = {data_json};")
```

`json.dumps` escapes for JSON, but does **not** escape `<`. The `embedded`
payload includes `runs`, which carry the **raw outputs each skill under test
produced** — frequently HTML, code, or markdown. As soon as any embedded
string contains `</script>` (or `</` in general, or `<!--`, or `<script`), the
browser's HTML tokenizer closes the inline `<script>` element early, before the
JS parser ever sees the string context. Two consequences:

- **Breakage:** `EMBEDDED_DATA` is never fully assigned, the viewer's JS throws,
  and the page renders blank — the reviewer sees nothing.
- **Injection:** output such as `</script>` becomes
  live DOM and executes.

For skill-creator this is a likely hit rather than a corner case: any skill
being evaluated that emits an HTML file — or code that merely mentions a script
tag — poisons its own results viewer.

### Fix

Escape `<`, `>`, `&`, U+2028, and U+2029 to their JSON unicode-escape forms
after `json.dumps` — the same HTML-safe-JSON technique used by Flask's
`htmlsafe_dumps` and Django's `json_script`. The escaped forms parse back to
the identical characters, so the embedded data is unchanged; the HTML tokenizer
simply never sees a `<` again. U+2028/U+2029 are included because
`EMBEDDED_DATA` is consumed as a JS **object literal** (not `JSON.parse`), where
those two separators were illegal in string literals before ES2019.

### Verification

Round-tripping a payload whose output contains a `</script>` breakout, an
`` injection, `A&B`, and literal U+2028/U+2029:

- before: serialized JSON contains a literal `</` and breaks the `<script>` block
- after: contains no literal `<`, `>`, `</`, or line separators, and the escaped
  string parses back equal to the original payload — data preserved exactly

Verified under both `ensure_ascii=True` (current) and `ensure_ascii=False`, so
it composes cleanly with #1040 (which changes the same line to
`ensure_ascii=False` for non-ASCII readability — a complementary fix).
`python -m py_compile` passes on the changed file.

### DIFF
--- a/skills/skill-creator/eval-viewer/generate_review.py
+++ b/skills/skill-creator/eval-viewer/generate_review.py
@@ -276,7 +276,22 @@ def generate_html(
     if benchmark:
         embedded["benchmark"] = benchmark
 
-    data_json = json.dumps(embedded)
+    # Escape the characters that are unsafe to embed in the viewer's <script>
+    # block. json.dumps does not escape "<", so an embedded string containing
+    # "</script>" (common, since runs carry the raw outputs a skill under test
+    # produced) would close the script element early — blanking the viewer and
+    # allowing markup injection. U+2028/U+2029 are valid in JSON but were illegal
+    # in JS string literals before ES2019, and EMBEDDED_DATA is consumed as a JS
+    # object literal. This mirrors Flask's htmlsafe_dumps / Django's json_script;
+    # each escape parses back to the original character, so the data is unchanged.
+    data_json = (
+        json.dumps(embedded)
+        .replace("<", "\\u003c")
+        .replace(">", "\\u003e")
+        .replace("&", "\\u0026")
+        .replace("\u2028", "\\u2028")
+        .replace("\u2029", "\\u2029")
+    )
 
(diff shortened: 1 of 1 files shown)


ihororlovskyi:
I hit this in normal skill-creator usage while evaluating a web/UI skill, so this is not only a theoretical XSS/hardening issue.

One generated output contained an HTML file with a literal `</script>`. The eval viewer then became unreadable / showed broken raw content, and Playwright reported `Invalid or unexpected token`.

A local workaround with `json.dumps(embedded).replace("</", "<\\/")` fixed the viewer immediately, but this PR’s broader escaping approach looks like the right upstream fix. After reinstalling `skill-creator`, the issue came back because the upstream version still had the unescaped `json.dumps(embedded)` path.

Issue #1394, opened 2026-07-05
skill-creator: eval-viewer escapeHtml is not attribute-safe and is applied inconsistently (display-path XSS)


griffithsbs:
### Summary

The skill-creator eval-viewer (`skills/skill-creator/eval-viewer/viewer.html`)
assembles HTML by string concatenation and assigns it with `innerHTML` in
`renderGrades` and `renderBenchmark`, relying on a helper `escapeHtml()` at each
interpolation point. Two gaps let attacker-influenced data reach the DOM as
markup:

1. **`escapeHtml()` is not attribute-safe.** It is implemented with the
   `textContent` → `innerHTML` round-trip, which escapes `<`, `>`, and `&` but
   **not** `"` or `'`. It is nonetheless used inside an HTML **attribute** — the
   per-assertion table renders grader evidence into a `title="..."` tooltip
   (`renderBenchmark`, the `<span ... title="Run N: <evidence>">` line). Evidence
   containing a double quote breaks out of the attribute and can attach an event
   handler.

2. **Several externally-sourced strings are interpolated raw**, with no
   `escapeHtml()` at all: the config label in the per-eval breakdown rows (while
   the *same* value is escaped in the table header), and `metadata.timestamp`,
   `metadata.evals_run`, `metadata.runs_per_configuration`, and the `delta.*`
   cells.

### Why this is reachable

`EMBEDDED_DATA` (from `benchmark.json` / `grading.json`) carries data that is
user-authored (assertion text, eval and skill names) or that quotes the output
of the skill under test (grader "evidence"). None of that is trusted markup.

### Reproduction

1. In a grading result, set an assertion's `evidence` to:

   ```
   x" onmouseover="alert(document.domain)
   ```

2. Generate the review and open the **Benchmark** tab.
3. Hover the ✓/✗ badge for that assertion in the per-assertion table. The
   `title="Run 1: x" onmouseover="alert(document.domain)">` markup closes the
   attribute early and the injected `onmouseover` handler fires.

### Fix

- Rewrite `escapeHtml()` to escape `&`, `<`, `>`, `"`, and `'` explicitly, so it
  is safe in both element-text and attribute contexts.
- Route the remaining raw externally-sourced string interpolations through
  `escapeHtml()`, leaving only locally-computed numbers interpolated directly.

The escaped forms render identically for normal data, so there is no visible
change.

### Scope / related

- This is the **display path**. The **embed path** — `generate_review.py`
[...]

PR #1395, opened 2026-07-05
fix(skill-creator): make eval-viewer HTML escaping attribute-safe

griffithsbs:
Fixes #1394.

### Problem

`viewer.html` assembles HTML by string concatenation and assigns it with
`innerHTML` in `renderGrades` and `renderBenchmark`, relying on `escapeHtml()`
at each interpolation. Two gaps let attacker-influenced data (grader "evidence"
and assertion text can quote a skill-under-test's output; eval and config names
are user-authored) reach the DOM as markup:

- **`escapeHtml()` was not attribute-safe.** It used the `textContent` →
  `innerHTML` round-trip, which escapes `<`, `>`, `&` but not quotes — yet it is
  used inside a `title="..."` attribute in the per-assertion table. Evidence
  containing a double quote breaks out of the attribute and can attach an event
  handler (e.g. `x" onmouseover="alert(document.domain)`).
- **Several externally-sourced strings were interpolated raw**: the config label
  in the per-eval breakdown rows (while the same value is escaped in the header),
  and `metadata.timestamp` / `metadata.evals_run` / `metadata.runs_per_configuration`
  / the `delta.*` cells.

### Fix

- Rewrite `escapeHtml()` to escape `&`, `<`, `>`, `"`, `'` explicitly, so it is
  safe in both element-text and attribute contexts.
- Route the remaining raw externally-sourced string interpolations through it, so
  every externally-sourced string rendered via `innerHTML` is escaped and only
  locally-computed numbers are interpolated directly.

The escaped forms render identically for normal data, so there is no visible
change.

### Scope / related

This is the **display path**. The **embed path** — `generate_review.py` injecting
`EMBEDDED_DATA` into the inline `<script>` without escaping `</`, which lets a
`</script>` in skill output break out of the script element (#1075) — is fixed
separately in #1393. The two together harden the viewer end to end; they touch
different files and do not conflict. The xlsx preview relies on SheetJS's own
cell-text escaping and is left as-is.

### DIFF
--- a/skills/skill-creator/eval-viewer/viewer.html
+++ b/skills/skill-creator/eval-viewer/viewer.html
@@ -1096,10 +1096,17 @@
       return "#";
     }
 
+    // Escape for embedding in HTML markup built by string concatenation. Covers
+    // quotes as well as angle brackets and ampersand, so the result is safe in
+    // attribute values (e.g. title="...") as well as element text — the
+    // textContent round-trip does not escape quotes and is unsafe in attributes.
     function escapeHtml(text) {
-      const div = document.createElement("div");
-      div.textContent = text;
-      return div.innerHTML;
+      return String(text ?? "")
+        .replace(/&/g, "&amp;")
+        .replace(/</g, "&lt;")
+        .replace(/>/g, "&gt;")
+        .replace(/"/g, "&quot;")
+        .replace(/'/g, "&#39;");
     }
 
     // ---- View switching ----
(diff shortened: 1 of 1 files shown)


Echolonius:
+1 to this. I independently hit the same `escapeHtml` attribute gap (filed #1398 a day after this, now closed as a duplicate of it — sorry for the noise). One data point in support: I verified the `textContent`→`innerHTML` round-trip leaves `"` and `'` unescaped, which is exploitable via the `title="..."` interpolations in `renderBenchmark`; the regex replacement here closes that, and this PR additionally escapes the raw `metadata.timestamp`/`evals_run`/`configLabel`/delta interpolations mine didn't touch. This version is the more complete fix.

PR #1398, opened 2026-07-06
fix(skill-creator): make eval-viewer escapeHtml attribute-safe

Echolonius:
Fixes #1394.

## The bug

`escapeHtml()` in `skills/skill-creator/eval-viewer/viewer.html` used the `textContent` → `innerHTML` round-trip, which escapes `<`, `>`, and `&` but **not** `"` or `'`. It's used inside an HTML attribute in `renderBenchmark`'s per-assertion table (`title="Run N: <evidence>"`), so grader evidence containing a double quote breaks out of the attribute and can attach an event handler:

```
evidence = `x" onmouseover="alert(document.domain)`
→ title="Run 1: x" onmouseover="alert(document.domain)"
```

`EMBEDDED_DATA` carries assertion text, eval/skill names, and grader "evidence" — none of it is trusted markup.

Several other externally-sourced strings were also interpolated raw with no `escapeHtml()` at all: `metadata.timestamp`, `metadata.evals_run`, `metadata.runs_per_configuration`, the `delta.*` cells, and the per-eval-breakdown `configLabel` (inconsistently — the *same* computed value was already escaped in the assertion-table header two call sites away, just not in the breakdown rows).

## The fix

- Rewrite `escapeHtml()` to explicitly escape `&`, `<`, `>`, `"`, and `'`, so it's safe in both element-text and attribute contexts.
- Route the remaining raw interpolations through `escapeHtml()`.

Verified the exact repro from the issue no longer breaks out of the attribute, and that escaping is transparent for normal data (browsers decode the entities back to the same display text, so there's no visible change).

Out of scope (per the issue): the embed path (`generate_review.py` → inline `<script>`, issue #1075) and the xlsx preview (relies on SheetJS's own escaping).

## Test plan
- [x] Reproduced the reported XSS locally against the old `escapeHtml`, confirmed the new version neutralizes it (double quote → `&quot;`, no attribute break-out)
- [x] Confirmed benign strings render identically before/after (entities decode back to the same visible text)
- [ ] No existing test harness for this standalone HTML viewer — happy to add one if maintainers want a specific format

### DIFF
--- a/skills/skill-creator/eval-viewer/viewer.html
+++ b/skills/skill-creator/eval-viewer/viewer.html
@@ -1097,9 +1097,12 @@
     }
 
     function escapeHtml(text) {
-      const div = document.createElement("div");
-      div.textContent = text;
-      return div.innerHTML;
+      return String(text)
+        .replace(/&/g, "&amp;")
+        .replace(/</g, "&lt;")
+        .replace(/>/g, "&gt;")
+        .replace(/"/g, "&quot;")
+        .replace(/'/g, "&#39;");
     }
 
     // ---- View switching ----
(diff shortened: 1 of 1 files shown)


griffithsbs:
Looks like a straight duplicate of #1395 : would you mind thumbing/commenting/updating that one instead?

Echolonius:
You're right, @griffithsbs — #1395 predates this and covers strictly more (your regex `escapeHtml` fixes the attribute gap I targeted, plus the unescaped interpolation sites in `renderBenchmark` I missed). Apologies for not checking open PRs before filing, and for the slow reply. Closing this in favor of #1395 — I've left a supporting note there.

PR #1399, opened 2026-07-06
fix(skill-creator): escape eval viewer attributes

ebbsanchez:
Fixes #1394.

## Summary
- make the eval viewer escaping helper encode quotes in addition to `&`, `<`, and `>`
- route benchmark metadata and config labels through the escaping helper before assigning generated markup via `innerHTML`
- apply the same escaping helper update to the trigger eval review asset

## Verification
- `git diff --check`
- local string harness with the issue payload `x" onmouseover="alert(document.domain)` confirmed the tooltip source keeps the quote as `&quot;`, so it does not terminate the `title` attribute

### DIFF
--- a/skills/skill-creator/assets/eval_review.html
+++ b/skills/skill-creator/assets/eval_review.html
@@ -103,9 +103,12 @@
     }
 
     function escapeHtml(text) {
-      const div = document.createElement('div');
-      div.textContent = text;
-      return div.innerHTML;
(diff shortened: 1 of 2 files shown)


ebbsanchez:
Hi, gentle follow-up after a couple of weeks. This is still ready on my side; happy to adjust or close if it is not aligned with the project direction.

Issue #1414, opened 2026-07-09
skill-creator: run_eval.py's nested claude -p subprocess inherits parent stdin, causing an intermittent stall


ajmccaus:
## Problem

`run_single_query()` in `skills/skill-creator/scripts/run_eval.py` spawns the nested eval `claude -p` via `subprocess.Popen(cmd, stdout=..., stderr=..., cwd=..., env=...)` without redirecting `stdin`. The child inherits the parent's stdin file descriptor.

When `run_eval.py`/`run_loop.py` itself runs nested inside a `claude` session (a supported, intended use — the file explicitly strips `CLAUDECODE` from `env` to allow this), the child process can block briefly waiting on/polling the inherited stdin before it gives up and proceeds, adding an observed ~3s stall per query. At `--num-workers` > 1 this multiplies across concurrent subprocesses and inflates wall-clock eval time without any corresponding signal.

## Repro

Run any eval from inside a `claude -p` or Claude Code session:

    python -m scripts.run_loop --eval-set eval.json --skill-path <skill> --model <m>

Observe each `run_single_query` call taking ~3s longer than the same call run from a plain (non-nested) shell.

## Fix

Single-line change — explicitly discard the child's stdin instead of inheriting it:

```diff
--- a/skills/skill-creator/scripts/run_eval.py
+++ b/skills/skill-creator/scripts/run_eval.py
@@ -84,6 +84,7 @@ def run_single_query(
 
         process = subprocess.Popen(
             cmd,
+            stdin=subprocess.DEVNULL,  # nested claude -p must not wait on/inherit our stdin
             stdout=subprocess.PIPE,
             stderr=subprocess.DEVNULL,
             cwd=project_root,
```

## Note for maintainers

Neither open PR #1323 nor #1298 (both reworking this same function for the 0%-recall bug) touches the Popen call — this fix is orthogonal and should be folded into whichever lands, or applied standalone if merged first.

---
*Disclosure: this issue was researched and written by Claude (Claude Code)*

PR #1426, opened 2026-07-10
skill-creator: redirect nested claude -p stdin to DEVNULL in run_eval.py

viditchess64:
`run_single_query()` spawns the nested eval `claude -p` without redirecting stdin, so the child inherits the parent's stdin. When the eval runs nested inside a Claude session (a supported use — the code already strips `CLAUDECODE` from env for exactly this), the child stalls ~3s waiting on the inherited stdin, multiplied across `--num-workers`.

One-line fix: pass `stdin=subprocess.DEVNULL` to the Popen call.

Fixes #1414

Note: open PRs #1323 and #1298 rework this same function but don't touch the Popen call — this change is orthogonal and can be folded into either if they land first.

### DIFF
--- a/skills/skill-creator/scripts/run_eval.py
+++ b/skills/skill-creator/scripts/run_eval.py
@@ -84,6 +84,7 @@ def run_single_query(
 
         process = subprocess.Popen(
             cmd,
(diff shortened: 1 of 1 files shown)


Fiskers:
Hello

Issue #1431, opened 2026-07-13
`user-invocable: true` is not accepted by the current skill validator


setoelkahfi:
Using Codex CLI:

<img width="1302" height="166" alt="Image" src="https://github.com/user-attachments/assets/24b33f71-cdaa-4398-b0d6-ded2f68f377e" />

PR #1434, opened 2026-07-14
Accept documented Claude Code frontmatter fields in the skill validator

maksmarcinkiewicz:
Fixes #1431

## Problem

`skills/skill-creator/scripts/quick_validate.py` rejects `SKILL.md` files that use documented Claude Code frontmatter fields such as `user-invocable: true`, because its allowlist only contains the six base Agent Skills spec fields (https://agentskills.io/specification). The same check also rejects `disable-model-invocation`, `argument-hint`, `model`, `hooks`, and the other extension fields documented at https://code.claude.com/docs/en/skills#skill-md-frontmatter — so a skill written exactly per the Claude Code docs fails validation.

## Change

The allowlist is now the union of two named sets: the Agent Skills spec fields and the documented Claude Code extensions, each with a source link. Unknown keys (e.g. typos like `descriptoin`) are still rejected, so the typo-catching value of the check is preserved. `package_skill.py` imports the same `validate_skill`, so packaging picks up the fix automatically.

## Verification

- A skill with `user-invocable: true`, `disable-model-invocation: true`, and `argument-hint` previously failed with `Unexpected key(s) in SKILL.md frontmatter`; now it validates.
- A skill with a misspelled key (`descriptoin`) is still rejected.
- All 17 skills in this repo validate with no regressions (the one pre-existing failure, `claude-api`'s over-long description, fails identically on `main`).

### DIFF
--- a/skills/skill-creator/scripts/quick_validate.py
+++ b/skills/skill-creator/scripts/quick_validate.py
@@ -39,11 +39,12 @@ def validate_skill(skill_path):
         return False, f"Invalid YAML in frontmatter: {e}"
 
     # Agent Skills spec fields (agentskills.io/specification)
-    # + documented Claude Code extensions (code.claude.com/docs/en/skills#skill-md-frontmatter)
+    # + documented Claude Code extensions (code.claude.com/docs/en/skills#frontmatter-reference)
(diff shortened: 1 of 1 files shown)


rr-krupesh-savaliya:
I think we need to add `when_to_use` as well, as per documented here: https://code.claude.com/docs/en/skills#frontmatter-reference

Issue #1417, opened 2026-07-10
Plan file hygiene / cleanup skill  sessions accumulate stray PLAN.md,  scratch files with no consolidation


halilxibrahim:
Over long running agentic sessions (e.g. Claude Code, or others AI ), each planning or research step tends to spawn its own
PLAN.md, NOTES.md, or scratch markdown file. Over time these accumulate in
the repo/workdir with no lifecycle: nothing merges them, prunes stale ones,
or distinguishes "still relevant" from "superseded by a later plan."

Proposing a skill that would:
- Detect existing planning/scratch .md files in the working directory
  before creating a new one
- Offer to merge/update an existing plan file instead of spawning a new one
- Flag or archive stale plan files once their associated task is completed

This would reduce repo clutter and make it easier to find the "current"
plan when returning to a project after a break.

Happy to take a first pass at implementing this myself if the idea sounds
useful would appreciate feedback on scope/approach before I start.

xg-gh-25:
This is a lifecycle-gap problem — planning artifacts accumulate with no consolidation or pruning logic.

**Core issue:**

Long-running sessions spawn PLAN.md, NOTES.md, or scratch files at each planning step, but nothing:
1. Detects that a plan already exists before creating a new one
2. Merges/updates an existing plan when context evolves
3. Archives or marks stale plans when their associated task completes
4. Distinguishes "current working plan" from "superseded historical plan"

The result: repos accumulate planning debris, and returning to a project after a break requires manually scanning timestamps/content to find the "real" plan.

**Production pattern we use:**

In SwarmAI's autonomous pipeline, we solve this with a **durable planning ledger** that lives *outside* the working directory but is session-accessible:

1. **Single canonical plan file per project**: `~/.swarm-ai/<project>/plan.md` (not in the repo). Each planning step *updates* this file (append or section-replace), never spawns a new one.

2. **Explicit lifecycle transitions**: When a task completes, the ledger records:
   ```markdown
   ## Task: Implement feature X
   Status: COMPLETED (2026-07-10)
   Archive: plan-archive/2026-07-10-feature-x.md
   ```
   The completed plan is moved to an archive directory, leaving the ledger clean for the next task.

3. **Session resume hook**: On `--resume`, the agent reads the ledger first. If a current plan exists, it surfaces that as context ("You have an open plan from 2026-07-09. Continue or start fresh?"). This prevents accidental plan-duplication when resuming.

**Alternate approach (lighter, in-repo):**

[...]

PR #1479, opened 2026-07-25
Add plan-file-hygiene skill (addresses #1417)

tonydzi:
Addresses #1417.

Credit where it's due: @halilxibrahim named the problem precisely (planning artifacts accumulate with no lifecycle), and @xg-gh-25's comment framed it as a lifecycle gap — this skill is built on that framing. @halilxibrahim, you mentioned wanting to take a first pass: happy to hand this off to you, fold it into your implementation, or iterate on it together — please treat this as a working draft toward your proposal, not a claim on it.

## What this adds

`skills/plan-file-hygiene/` — a single SKILL.md (plus bundled Apache-2.0 LICENSE.txt) with four rules and a reporting contract:

1. **Discover before you create** — search for existing `PLAN*/NOTES*/TODO*/scratch*` markdown and classify (current / superseded / stale / unrelated) before spawning a new file.
2. **Merge, don't spawn; append, don't regenerate** — update the existing plan via re-read-then-targeted-edit; never rewrite a shared plan file from session memory. Includes concrete "you are about to clobber" warning signs (writing without reading this turn; output smaller than the existing file).
3. **One current plan; supersede explicitly** — restructures replace the old plan with a dated supersede marker, never a `PLAN-v2.md` sibling.
4. **Close the lifecycle** — mark done + archive on completion, staleness sweep for cluttered directories, and never silently delete another writer's (or uncommitted) file.

Plus a required hygiene report, a worked example, and edge cases (concurrent writers, multi-task directories, ephemeral scratch).

## Where it comes from

The rules are distilled from a production multi-writer setup (multiple agent sessions and humans sharing plan files). Repeated plan-file losses there were traced through file version history to exactly one habit — regenerating the whole plan from what the current session remembered, which silently deleted other writers' sections — not to sync races, which were the initial suspect. The discipline in this skill is the fix that eliminated that failure class; the "shrinking plan file is a red flag" heuristic comes straight from that forensics.

## Testing

- Frontmatter validated against the agentskills.io spec constraints (name/dir match, charset, description length, body <500 lines).
- Followed literally by a separate agent session against a fixture directory (a current plan with open checkboxes, overlapping scratch notes from the same task, a month-old finished plan, and someone else's untracked TODO file). The resulting actions matched the intended decisions, including leaving the untracked file in place and flagging it in the report. The ambiguities that agent reported (which file wins when notes overlap the plan; shrink-warning vs. legitimate consolidation; targeted-edit vs. restructure threshold; act-vs-propose on archiving) were fixed in the text.
- Two adversarial review passes by a second, independent model; its findings (read-then-rewrite race caveat, ownership gate on deletion and superseding, supersede-marker ordering on interruption) are folded in.

## Disclosure

Architecture, the underlying incident forensics, and validation are mine; the text was drafted in pair with Claude.

### DIFF
--- a/skills/plan-file-hygiene/LICENSE.txt
+++ b/skills/plan-file-hygiene/LICENSE.txt
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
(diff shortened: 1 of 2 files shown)


halilxibrahim:
Thanks for putting this together and for the attribution. @Palo-Alto-AI-Research-Lab 

The implementation looks solid. When I have some time, I’d be happy to contribute a small follow-up to the PR, especially around post edit verification: checking the diff after an update.

tonydzi:
Thanks @halilxibrahim — that would be great, and post-edit verification is exactly the right next layer: the current skill checks hygiene *before* an edit lands, so diffing *after* an update closes the loop. Feel free to push to this branch or open a follow-up PR referencing it, whichever is easier for you — happy to review same-day either way.

Issue #1469, opened 2026-07-24
pdf skill: extract_form_field_info.py corrupts/crashes on comboboxes whose /Opt entries are bare strings


musharna:
## Summary

`skills/pdf/scripts/extract_form_field_info.py` assumes every choice-field option (`/Opt`) is a `[export_value, display_text]` pair. Per ISO 32000-1 §12.7.4.4, an `/Opt` array entry may instead be a **single text string** (used as both the export value and the display text). On such fields the script silently produces corrupted output, and crashes outright when any option is a single character.

## Buggy code (`skills/pdf/scripts/extract_form_field_info.py`, lines 35-41)

```python
    elif ft == "/Ch":
        field_dict["type"] = "choice"
        states = field.get("/_States_", [])
        field_dict["choice_options"] = [{
            "value": state[0],
            "text": state[1],
        } for state in states]
```

`states` comes from pypdf's `/_States_`, which for `/Ch` fields is the raw `/Opt` array verbatim (`pypdf/_doc_common.py`: `retval[key]["/_States_"] = obj["/Opt"]`). When `/Opt` holds bare strings, `state[0]`/`state[1]` index into the string instead of a pair.

## Reproduction

Create a combobox whose `/Opt` is bare strings, then run the script (pypdf 6.13.1):

```python
from pypdf import PdfWriter
from pypdf.generic import (DictionaryObject, ArrayObject, NameObject,
    TextStringObject, NumberObject)
w = PdfWriter(); page = w.add_blank_page(width=300, height=300)
f = DictionaryObject()
f[NameObject("/FT")] = NameObject("/Ch")
f[NameObject("/T")]  = TextStringObject("colors")
f[NameObject("/Ff")] = NumberObject(1 << 17)  # Combo
f[NameObject("/Opt")] = ArrayObject([TextStringObject(s) for s in ("Red","Green","Blue")])
f[NameObject("/Subtype")] = NameObject("/Widget")
f[NameObject("/Rect")] = ArrayObject([NumberObject(x) for x in (50,50,250,80)])
f[NameObject("/P")] = page.indirect_reference
ref = w._add_object(f); page[NameObject("/Annots")] = ArrayObject([ref])
acro = DictionaryObject(); acro[NameObject("/Fields")] = ArrayObject([ref])
w._root_object[NameObject("/AcroForm")] = w._add_object(acro)
w.write(open("bare_opt.pdf","wb"))
```

```
[...]

PR #1501, opened 2026-07-28
pdf: preserve bare string choice options

tomatotomata:
## Summary\n- normalize PDF choice-field /Opt entries that are bare strings\n- preserve existing [value, label] pair handling\n- add regression tests for both representations\n\nFixes #1469\n\nValidation: python -m unittest discover -s scripts -p 'test_*.py' -v, python -m py_compile scripts/extract_form_field_info.py scripts/test_extract_form_field_info.py, git diff --check\n\nAssisted-by: Codex


tomatotomata:
Closing this duplicate so the existing upstream PR #1474 remains the canonical fix for #1469. The implementation overlaps; its additional regression tests can be revisited if maintainers request them.

PR #1532, opened 2026-08-06
Update claude-api skill: Managed Agents August launch wave

cj-ant:
## What this does

Updates the `claude-api` skill to match the current upstream source of the skill content: the Managed Agents features that shipped in the August 5 launch wave and a batch of corrections to behavior the skill already documented but got wrong.

### Managed Agents: August 5 launch wave

| Feature | Shape | Sharp edge |
|---|---|---|
| Session budgets | `budget: {type: "limit", max_list_cost: {amount, currency}}` at session create; `budget_reached` stop reason; `session.usage` event | `amount` is an integer minor-unit string (`"2500"` = $25.00, decimals rejected). Create-only: a cap can be changed or removed later, never added. Removal is one-way. At the cap only settle events are accepted — a `user.message` is a 400, and a `user.interrupt` is accepted and silently ignored. |
| Deployment budgets | Same `budget` object on deployment create/update, copied onto each fired session | Opposite update semantics from sessions: clearable with `null` *and* re-addable. Changes apply from the next fire. |
| Inference geo pinning | `inference_geo` (`"us"` \| `"global"`) on the agent's `model` object | Nested in `model`, never top-level (the reverse of the Messages API). Validated at save, session create, and every turn — never grandfathered. Rosters must be geo-uniform. A per-session `model` override that omits it clears the pin, because overrides replace the object in full. |
| Skills from a GitHub repository | Root `.claude/skills/<name>/SKILL.md` discovered at session start when a repo is mounted | Cloud sandboxes only. Scanned once, at session start — commits pushed mid-session are not picked up. Repository skills are agent instructions: anyone who can commit to the repo is inside the trust boundary. |
| Advisor | `{type: "advisor", model}` multiagent roster entry; reserved `anthropic.advisor` thread | Advice delivery is not guaranteed to precede the advisor thread's idle/terminated events. Redacted advisors deliver `[{"type": "redacted"}]` to the client while the agent reads the advice server-side. Advisor threads are exempt from the 25-thread limit. |

Supporting updates: multiagent "when to use it" guidance (start with a `{"type": "self"}` roster, then move reading-heavy work to a cheaper worker agent), the deployments `update` endpoint, and the advisor tool's `max_uses` / `max_tokens` / `caching` options and error result variant on the Messages API side.

### Corrections to what the skill already said

- **Agent versions are sequential integers** starting at 1, not numeric timestamps.
- **Files API uploads take a `purpose` parameter** in the managed-agents examples.
- **`stop_details` refusal categories are an open set** (`"reasoning_extraction"`, `"frontier_llm"`, …), not just `"cyber" | "bio" | null`.
- **Sonnet 5 belongs on the prefill-removed list.**
- **`vault_ids` is create-only on sessions** — the update param exists in the SDK but the API rejects it.
- **The multi-agent docs page moved** to `managed-agents/multiagent-orchestration.md`.


cj-ant:
@rlancemartin @ericharmeling — same shape as the previous claude-api content syncs (#1463, #1476): the August 5 Managed Agents wave plus a batch of accuracy corrections. Ready for review when you have a moment.

nikblanchet (approved):
:shipit:

ericharmeling (approved):

Issue #881, opened 2026-04-07
claude-api skill description truncated in Claude Code listings


azarovalex:
The `claude-api` skill's `description` field in frontmatter exceeds 250 characters, which causes Claude Code to truncate it in the skill listing.

What Claude Code sees in context:

```
claude-api: Build apps with the Claude API or Anthropic SDK.
TRIGGER when: code imports `anthropic`/`@anthropic-ai/sdk`/`claude_agent_sdk`, or user asks to use Claude API, Anthropic SDKs, or Agent SDK.
DO NOT TRIGGER when: code imports `openai`/other AI SDK, ge…
```

The `DO NOT TRIGGER when` conditions are cut off, which means Claude may incorrectly invoke this skill when working with non-Anthropic AI SDKs.

Per the [Claude Code docs on skills](https://docs.anthropic.com/en/docs/claude-code/skills#skill-files), descriptions longer than 250 characters are truncated in skill listings to reduce context usage. The fix would be to shorten the description to under 250 chars, front-loading the key use case and keeping the exclusion conditions concise.

azarovalex:
I believe this problem also applies to some other native skills like `/loop` and `/update-config`

ChristopheChirat:
Adding a current, stricter data point on this skill's `description` length.

Since this issue was filed, the `claude-api` `description` was rewritten and is now **1068 characters** (the YAML-parsed value of `skills/claude-api/SKILL.md` on `main`). Beyond the 250-char *listing truncation* described here, it now also crosses a **hard 1024-character limit**, with a more severe consequence in **GitHub Copilot CLI** (observed v1.0.61–1.0.62):

```
The following skills failed to load:
✗ ...\anthropic-agent-skills\document-skills\skills\claude-api\SKILL.md: Skill ...
```

i.e. the skill is **rejected entirely and does not load** — not merely truncated. Copilot CLI ships a native validator (`runtime.node`) that enforces a max `description` of 1024 chars (`"description ... exceeds maximum"`), and the Agent Skills standard documents the same 1024 ceiling.

**Repro** (provider-agnostic):

```python
import yaml
d = open("skills/claude-api/SKILL.md", encoding="utf-8").read()
desc = yaml.safe_load(d.split("---", 2)[1])["description"]
print(len(desc))  # -> 1068  (44 over the 1024 limit)
```

[...]

PR #1538, opened 2026-08-09
fix: bring two skills back under the Agent Skills spec

bechor25:
Two skills in this repository fail `skills-ref validate` against the spec this repository is the reference implementation for.

---

## 1. `template/SKILL.md` — name does not match its directory

```yaml
name: template-skill     # parent directory is  template/
```

> The required `name` field: … **Must match the parent directory name**
> — <https://agentskills.io/specification#name-field>

**Which side is wrong:** the directory. `README.md:65` already advertises it as *"the **template-skill** in this repository"*, and `git grep "template/"` returns nothing — no file references the path. So renaming the directory to `template-skill/` keeps the name the README promises and breaks no links.

---

## 2. `skills/claude-api/SKILL.md` — description is 1068 characters

> The required `description` field: **Must be 1-1024 characters**
> — <https://agentskills.io/specification#description-field>

44 characters had to go. Rather than rewrite any trigger, I removed 48 characters of redundancy:

| removed | why it is redundant |
| --- | --- |
| `langchain_openai\|` | inside a `grep -E` alternation, the earlier `openai` branch already matches every `langchain_openai` occurrence — the alternative can never fire |
| `Claude, Anthropic, ` | the clause immediately before it already reads *"the prompt names Claude/Anthropic"* |
| ` in any form` | the parenthetical that follows is the enumeration of the forms |

**No trigger token, model glob, or provider name was dropped.** `Fable`, `Opus`, `Sonnet`, `Haiku`, `` `anthropic` ``, `` `@anthropic-ai` ``, `` `claude-*` ``, `` `us.anthropic.*` ``, `` `[1m]` ``, and every entry in both provider lists are untouched.

Result: **1020 characters.**

If you would rather spend the 44 characters somewhere else in that description, say so and I will move the cut — the only thing that matters is landing under the cap.

---

Found with [AgentCompass](https://github.com/YoavLax/agent-compass), an offline static analyzer for AI-agent repo readiness. Both findings were verified by hand against the spec before opening.


bechor25:
Still `MERGEABLE` against `main`. Adding a note in case it makes review cheaper:

**The two fixes are independent and I'm happy to split them.** If only one is wanted, say which and I'll drop the other.

1. `template/SKILL.md` declares `name: template-skill` in a directory named `template`. I renamed the directory rather than the field, because `README.md:65` already advertises it as "the **template-skill** in this repository" and `git grep "template/"` returns nothing.

2. `skills/claude-api/SKILL.md` has a 1068-character description against the 1024 cap. I removed 48 characters I could argue were redundant rather than rewriting any trigger — `langchain_openai|` (the earlier `openai` branch in the same `grep -E` alternation already matches it), `Claude, Anthropic, ` (the preceding clause names both), and ` in any form` (the parenthetical enumerates them).

That second one is a judgement call about someone else's carefully tuned trigger text, so if you'd rather spend those 44 characters elsewhere, tell me where and I'll move the cut. The only thing that matters is landing under the cap.

Both are reproducible with `skills-ref validate`.

zynovex-support:
**Additional data point supporting fix #2 (the 1068-char `claude-api` description): GitHub Copilot CLI hard-fails to load the skill entirely — it's not just a truncation issue.**

Because the [Agent Skills spec](https://agentskills.io/specification#description-field) mandates `description` be 1–1024 characters, GitHub Copilot CLI enforces this as a hard validation error (`z.string().max(1024, "Skill description must be at most 1024 characters")`). The oversized description doesn't get truncated — the whole skill is **rejected and dropped from the registry**.

Reproduced on **GitHub Copilot CLI 1.0.79** with the current upstream `claude-api/SKILL.md` (description = 1068 chars, 44 over the limit):


$ copilot skill list
...
✖ The following skills failed to load: • .../anthropic-agent-skills/document-skills/skills/claude-api/SKILL.md: Skill description must be at most 1024 characters


So the impact is broader than the Claude Code truncation noted in #881: on Copilot CLI the `claude-api` skill is silently unavailable until this PR lands. This +2/−2 change fixes it cleanly. 🙏

Issue #1552, opened 2026-08-12
skill-creator: description optimizer reports ~0% recall for skills that actually do trigger


HiroSnow0413:
# skill-creator: description optimizer reports ~0% recall for skills that actually do trigger

Two bugs in `example-skills/skills/skill-creator/scripts/run_eval.py` cause trigger evaluation to record a skill as "not triggered" when it demonstrably *is* triggered. Both fail in the same direction — they only ever turn a real trigger into a miss — so `run_loop.py` optimizes against a false signal and rewrites descriptions that were already fine.

## Symptom

Running the description optimization loop on a freshly written skill:

```
Train: 18/36 correct, precision=100% recall=0% accuracy=50%
Test : 13/24 correct, precision=100% recall=8% accuracy=54%
```

Every positive query scored `rate=0/3`. Capturing the raw `claude -p` stream for one of those queries and replaying it through `run_single_query`'s own detection logic returns `True` — the model called `Skill` with the expected `clean_name`. So the queries trigger; the harness fails to record it.

## Bug 1: the drained tail is never parsed

`run_single_query` reads the subprocess stream incrementally. When `process.poll()` shows the child has exited, it reads the remainder into `buffer` and then breaks out of the loop — but parsing only happens in the branch below, on freshly `os.read`-ed chunks. Whatever was in that final read is discarded unparsed, and the function falls through to `return triggered` (still `False`).

```python
while time.time() - start_time < timeout:
    if process.poll() is not None:
        remaining = process.stdout.read()
        if remaining:
            buffer += remaining.decode("utf-8", errors="replace")
        break            # <-- buffer never parsed
    ...
```

This hits whenever the child's output lands in one go, which is the common case for a query the model answers in a single turn.

## Bug 2: parallel workers share one commands directory

All workers write their temporary skill copy into the same `<project_root>/.claude/commands/`, and each `claude -p` runs with `cwd=project_root`. With `--num-workers N`, every run therefore sees its own copy *plus* the N-1 copies belonging to the other in-flight runs — N near-identical descriptions competing for the same query. Detection requires the run's own `clean_name`, so a run has roughly a 1/N chance of being credited.

That is exactly the observed `recall=8%` at the default `--num-workers 10`. This one is nastier than bug 1 because it yields a plausible-looking nonzero number rather than an obvious zero, and it scales with worker count, so it reads as a description quality problem.

## Fix

Both are addressed in `run_eval.py`:

[...]

Issue #1487, opened 2026-07-27
`claude-api` skill eagerly injects ~156k tokens, exhausting the context window in a single tool call


DaKev:
# `claude-api` skill eagerly injects ~156k tokens, exhausting the context window in a single tool call

## Environment

| | |
|---|---|
| Claude Code | 2.1.220 |
| Skill | `claude-api` (bundled at `%LOCALAPPDATA%\Temp\claude\bundled-skills\2.1.220\<hash>\claude-api`; also published via `anthropic-agent-skills` → `github.com/anthropics/skills`) |
| Model | `claude-opus-5` (default context) |
| OS | Windows 11 (10.0.22631), Git Bash |
| Node | v24.14.0 |

## Summary

Invoking `Skill(claude-api)` injects a single message of **624,401 characters (~156k tokens)** — the concatenation of `SKILL.md` plus **31 reference documents**. On a default-context model this consumes the entire window, and the very next API call fails with `Prompt is too long`. The session is unrecoverable.

This is not a slow degradation. It is one tool call from a fresh session to a dead session.

## Reproduction

1. Start a new session on `claude-opus-5` (default context).
2. Prompt with a request to edit a Markdown file that mentions model names in prose, e.g.:

   > Opus 5 is now available and is more capable than Fable 5. Please update the model list in our global CLAUDE.md accordingly.

3. The model reads `~/.claude/CLAUDE.md` (~1 KB), then invokes `Skill(claude-api)` per the skill's trigger.
4. Next assistant turn: `Prompt is too long`.

The task is editing a Markdown table in a config file. No code, no API calls, no SDK usage anywhere in scope.

*(The prompt above is minimized from the one that produced the transcript below; it retains the tokens that match the trigger — `Opus`, `Fable`, `CLAUDE.md`.)*

### Transcript evidence

| Entry | Type | Size |
|---|---|---|
| 6 | user prompt | 683 B |
| 13 | `Read` of a ~1 KB Markdown file | ~1 KB |
| 14 | `tool_use` / `Skill` | 1.2 KB |
| **17** | **skill payload** | **624,401 chars (~156k tokens)** |
| 19 | assistant | `Prompt is too long` |

## Expected vs. actual

**Expected:** the skill loads its entry document and reads references on demand, as the progressive-disclosure model for skills implies.

**Actual:** 32 documents arrive at once, before the model knows which — if any — are relevant.

### 1. Over half the payload is provably irrelevant to any given task

Composition of the 624,401 characters:

[...]

PR #1578, opened 2026-08-15
Restructure claude-api for progressive disclosure; guard against the #1487 failure class

Swastikbhat-lab:
## Restructure `claude-api` for progressive disclosure; guard against the #1487 failure class

Fixes the structural root cause of **#1487**: the `claude-api` skill — the repo's flagship skill, bundled with Claude Code — injected ~156k tokens in a single tool call, exhausting the context window on default-context models. Any mention of a model name (e.g. editing a `CLAUDE.md`) could trigger it and kill the session.

### What changed

**1. `skills/claude-api/` — SKILL.md is now a compact index (progressive disclosure)**

All 18 inline quick-reference sections (models, thinking/effort, caching, compaction, fast-mode, task budgets, provider clients, context editing, server tools, tool-use patterns, common pitfalls, …) were moved byte-exact into `references/*.md`, read on demand per the spec's three-level loading model ([agentskills.io/specification](https://agentskills.io/specification)). The trigger was narrowed: prose-only mentions of model names no longer fire the skill — it now requires a real API/SDK signal — and the anti-judgement "don't skip because it looks like a one-liner" language was removed. The non-Anthropic-provider guard is preserved.

**2. `skills/skill-creator/scripts/quick_validate.py` — upgraded into a progressive-disclosure validator**

The validator now enforces the spec budgets (warnings by default, errors with `--strict`), verifies every file the SKILL.md index references actually exists (a broken index is a broken skill), and prints the eager-payload metric (#1487's number) on every run. Also fixed a real bug: it crashed on UTF-8 content on Windows (cp1252). The `validate_skill()` contract used by `package_skill.py` is unchanged.

**3. `skills/skill-creator/` — the meta-skill practices what it preaches**

The guard honestly flagged skill-creator at 8,156 tokens. Its two heaviest procedural sections moved to `references/` with explicit pointers.

**4. `scripts/run_eval.py` — trigger-evaluation bugs from #1552/#1559 fixed** (drained tail never parsed, parallel workers sharing one commands dir, early return on non-skill tool calls).

**5. CI — `.github/workflows/ci.yml`** runs pytest + the strict validation sweep on every push, via the new `scripts/validate_all.py` sweep.

### Before → after

**`skills/claude-api/SKILL.md`** (the thing injected on trigger):

| Metric | Before | After | Spec |
|---|---|---|---|
| Size | 72,531 chars | **19,067 chars** (−74%) | < 20 KB |
| Estimated tokens | ~18,132 | **~4,766** (−74%) | < 5,000 |
| Lines | 547 | **235** (−57%) | < 500 |
| Description | 1,068 chars (over limit) | **883** | ≤ 1,024 |

**`skills/skill-creator/SKILL.md`** (flagged by the new guard, then slimmed):

| Metric | Before | After | Spec |
|---|---|---|---|
| Estimated tokens | ~8,156 | **~4,991** (−39%) | < 5,000 |
| Lines | 483 | **275** | < 500 |

### Test results

```
44 passed
```

- **Unit tests** for every validator check: frontmatter, name rules, description limits, reference integrity (incl. `{lang}` placeholders and globs), line/token budgets, payload metrics, UTF-8 robustness
- **Regression tests** prove the *old* claude-api structure fails strict validation and the *new* one passes — the #1487 guard
- **run_eval tests** (10) cover all three bug classes, written first and failing 8/10 against the old code
- **Sweep tests** cover the CI gate: pass/fail exit codes, JSON output, non-skill dirs ignored
- Full repo sweep: **all 17 skills pass strict-mode validation**

### Notes

- The eager injection itself happens in the Claude Code runtime, so this PR fixes the skill's structure — a loader that injects only SKILL.md now lands at ~19 KB instead of ~72 KB, and the skill no longer *is* a 156k-token payload. Closing the loop is the new CI gate, which fails if an oversized SKILL.md ever regresses.
- Content-drift verified: all 20 reference files are byte-exact extractions of their source sections (whitespace-normalized); the two intentionally compressed files (`managed-agents.md`, `agent-design.md`) preserve all facts, with the moved content verified present in the skill.
- Commits are atomic per issue: `#1552/#1559` run_eval fixes in their own commit, `#1487` restructure in another.


Swastikbhat-lab:
Closing — will reopen with smaller, focused changes.

PR #1591, opened 2026-08-16
fix(skill-creator): add encoding=utf-8 to all text file operations

### LINKED ISSUE #1304: Windows: eval/optimize harness broken — cp1252 read_text() crash + nested claude -p WinError 10038 silently zeroes all results
## Environment
- OS: Windows 11
- Python 3.14.4
- `skill-creator` from `anthropics/skills` (example-skills plugin), run via `python -m scripts.run_loop`

Two Windows bugs make the eval / description-optimization harness unusable; the second is the dangerous one because it **fails silently and reports plausible-looking but meaningless metrics**.

## Bug 1 — `parse_skill_md` crashes on any non-cp1252 SKILL.md (blocking)

`scripts/utils.py:9` reads SKILL.md with no explicit encoding:

```python
content = (skill_path / "SKILL.md").read_text()
```

On Windows the default text encoding is cp1252, so any SKILL.md containing UTF-8 beyond cp1252 — emoji (✅ 🟡), arrows (→), em-dashes (—), middots (·) — aborts the whole run:

```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 2363: character maps to <undefined>
  File ".../scripts/utils.py", line 9, in parse_skill_md
    content = (skill_path / "SKILL.md").read_text()
```

Same latent issue at many other sites: `scripts/quick_validate.py:22`, `generate_report.py:314`, and numerous `open()` / `write_text()` calls in `aggregate_benchmark.py`, `run_eval.py`, `run_loop.py`, `improve_description.py`.

**Fix:** pass `encoding="utf-8"` to every `read_text()` / `write_text()` / `open()` in `scripts/`. Minimal blocker fix:

```python
content = (skill_path / "SKILL.md").read_text(encoding="utf-8")
```

Workaround: run under `PYTHONUTF8=1`.

## Bug 2 — nested `claude -p` fails on Windows and the failure is scored as "no trigger" (critical)

After working around Bug 1 with `PYTHONUTF8=1`, `run_loop` runs to exit 0 — but **every** query (should-trigger and should-not-trigger alike) returns `trigger_rate: 0.0`, with stderr flooded (~40 per iteration) by:

```
Warning: query failed: [WinError 10038] An operation was attempted on something that is not a socket
```

`run_eval.py` runs each triggering check by spawning a nested `claude -p` via `ProcessPoolExecutor` + `subprocess.Popen` (≈ lines 16 / 85). On Windows every subprocess call fails with WinError 10038, so no triggering verdict is ever obtained.

The harness treats a failed query as "did not trigger," so should-not-trigger queries pass by default and the run reports a confident-looking result (`accuracy 50%, precision 100%, recall 0%`) plus a `best_description` — when in fact nothing was measured. This is easy to mistake for a real benchmark.

**Suggested fix:** abort (or loudly fail the iteration) when a high fraction of queries error, instead of scoring errored queries as non-triggers. Separately, the nested-`claude -p` subprocess path needs a Windows-compatible execution strategy.

**Repro:**
```
PYTHONUTF8=1 python -m scripts.run_loop --eval-set eval.json --skill-path <skill> --model claude-opus-4-8 --num-workers 4
```

Xsidz:
## What

On Windows, the default text encoding is cp1252. Any SKILL.md or output file containing UTF-8 beyond cp1252 (emoji ✅ 🟡, arrows →, em-dashes —) raises `UnicodeDecodeError` / `UnicodeEncodeError`. The immediate crash is in `parse_skill_md` (utils.py) which blocks every script. The write in `run_loop.py` is the worst case — it crashes after the expensive eval has already run, losing all results.

Adds `encoding="utf-8"` to every `read_text` / `write_text` / `open` for text files in:

| File | Sites |
|---|---|
| `utils.py` | `parse_skill_md` read (blocking crash) |
| `quick_validate.py` | SKILL.md read |
| `run_eval.py` | command file write |
| `run_loop.py` | eval set read, live/final report writes (×4), results.json write |
| `generate_report.py` | HTML output write |
| `improve_description.py` | log write, eval results read, history read |

`aggregate_benchmark.py` has the same issue but is currently touched by open PR #1590; adding encoding there would create a merge conflict and is left for that PR.

## Workaround until merged

`PYTHONUTF8=1 python -m scripts.run_loop ...`

## Test plan

- [ ] A SKILL.md containing emoji (e.g. ✅) loads without error on Windows (or with `PYTHONUTF8=0` forced)
- [ ] The HTML report is written after a successful run on a non-UTF-8 locale

Fixes #1304

### DIFF
--- a/skills/skill-creator/eval-viewer/generate_review.py
+++ b/skills/skill-creator/eval-viewer/generate_review.py
@@ -91,7 +91,7 @@ def build_run(root: Path, run_dir: Path) -> dict | None:
     for candidate in [run_dir / "eval_metadata.json", run_dir.parent / "eval_metadata.json"]:
         if candidate.exists():
             try:
-                metadata = json.loads(candidate.read_text())
+                metadata = json.loads(candidate.read_text(encoding="utf-8"))
                 prompt = metadata.get("prompt", "")
                 eval_id = metadata.get("eval_id")
             except (json.JSONDecodeError, OSError):
@@ -104,7 +104,7 @@ def build_run(root: Path, run_dir: Path) -> dict | None:
         for candidate in [run_dir / "transcript.md", run_dir / "outputs" / "transcript.md"]:
             if candidate.exists():
                 try:
-                    text = candidate.read_text()
+                    text = candidate.read_text(encoding="utf-8")
                     match = re.search(r"## Eval Prompt\n\n([\s\S]*?)(?=\n##|$)", text)
                     if match:
                         prompt = match.group(1).strip()
@@ -131,7 +131,7 @@ def build_run(root: Path, run_dir: Path) -> dict | None:
     for candidate in [run_dir / "grading.json", run_dir.parent / "grading.json"]:
         if candidate.exists():
             try:
-                grading = json.loads(candidate.read_text())
+                grading = json.loads(candidate.read_text(encoding="utf-8"))
             except (json.JSONDecodeError, OSError):
                 pass
             if grading:
@@ -153,7 +153,7 @@ def embed_file(path: Path) -> dict:
 
     if ext in TEXT_EXTENSIONS:
         try:
-            content = path.read_text(errors="replace")
+            content = path.read_text(encoding="utf-8", errors="replace")
         except OSError:
             content = "(Error reading file)"
         return {
@@ -222,7 +222,7 @@ def load_previous_iteration(workspace: Path) -> dict[str, dict]:
     feedback_path = workspace / "feedback.json"
     if feedback_path.exists():
         try:
-            data = json.loads(feedback_path.read_text())
+            data = json.loads(feedback_path.read_text(encoding="utf-8"))
             feedback_map = {
                 r["run_id"]: r["feedback"]
                 for r in data.get("reviews", [])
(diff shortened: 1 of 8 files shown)


98zc5g5jyw-arch:
## Babysit review — correct, partially superseded

Adding `encoding="utf-8"` to all text operations is the right and necessary fix (Chinese environments crash otherwise). Note: this exact set of files is also touched by #1534 and #1630, and `generate_review.py`/`aggregate_benchmark.py` are missing from the batch — coordinate so only one utf-8 PR lands and the two stragglers get covered.

98zc5g5jyw-arch (approved):
## Hermes Review

**Verdict: Approve** — explicit UTF-8 encoding is correct and low-risk.

### Verified
- `Path.read_text()/write_text()` without `encoding=` uses the platform default (cp1252 on Windows) — a real crash/mojibake source for CJK and emoji content. Explicit `utf-8` is the right fix across all six files; no behavior regression on POSIX.

### Suggestion
- Duplication: #1596 (same 6 files + claude-api trim) and #1534 (ppcvote) also add UTF-8. Please coordinate so this lands once; the smallest clean diff is preferred.

Xsidz:
Pushed 735cdc6 which adds `encoding="utf-8"` to the two files you flagged: `aggregate_benchmark.py` (all `open()` calls) and `eval-viewer/generate_review.py` (all `read_text()` calls).

On the overlap with #1534, #1596, and #1630 — those are from other contributors so I can't close them. Up to maintainers which one lands.

tyeth (approved):
If only they didn't give everyone at Anthropic a Mac, so someone might merge some of these Windows bugs + PRs.
It's also the case that their software usually works around the problem, merely costing the user more tokens...
Related: https://github.com/anthropics/claude-plugins-official/issues/1431

PR #1593, opened 2026-08-16
fix(webapp-testing): kill the server's process group so servers aren't orphaned

romansrepairsltd-coder:
## Problem

`scripts/with_server.py` launches each server with `shell=True` and stops it with `process.terminate()`. That signals the **shell**, not the server the shell spawned. For any command of the documented multi-server form:

```bash
python scripts/with_server.py \
  --server "cd backend && python server.py" --port 3000 \
  --server "cd frontend && npm run dev" --port 5173 \
  -- python test.py
```

the shell dies and the real server is orphaned, still bound to its port. The script prints `All servers stopped` while the process is very much alive, and the next run fails with `EADDRINUSE` / `Address already in use`.

## Reproduction

```bash
python scripts/with_server.py --server "cd site && python3 -m http.server 8917" --port 8917 -- true
ss -ltn | grep 8917   # still listening after "All servers stopped"
```

## Fix

- `start_new_session=True` puts the shell and everything it spawns into one process group.
- Cleanup signals that group with `os.killpg` — `SIGTERM` first, `SIGKILL` after the existing 5s timeout.
- `ProcessLookupError` is tolerated on both paths, for a server that already exited on its own.

Behaviour is unchanged when the server is a direct child; this only extends the reach of the existing terminate/kill escalation.

## Verification

Reproduced the orphan on Linux (Ubuntu 24.04, Python 3.12), confirmed the port stays bound after cleanup, applied the patch, and confirmed the port is released. Checked with a page whose content is set by JS, so the run also exercises the documented `networkidle` path via Playwright before teardown.

### DIFF
--- a/skills/webapp-testing/scripts/with_server.py
+++ b/skills/webapp-testing/scripts/with_server.py
@@ -18,6 +18,8 @@ import subprocess
 import socket
 import time
 import sys
+import os
+import signal
 import argparse
 
 def is_server_ready(port, timeout=30):
@@ -66,11 +68,15 @@ def main():
             print(f"Starting server {i+1}/{len(servers)}: {server['cmd']}")
 
             # Use shell=True to support commands with cd and &&
+            # start_new_session puts the shell and everything it spawns into one process
+            # group, so cleanup can signal the whole group. Without it, terminate() reaches
+            # only the shell and the real server is orphaned, still holding its port.
             process = subprocess.Popen(
                 server['cmd'],
                 shell=True,
                 stdout=subprocess.PIPE,
-                stderr=subprocess.PIPE
+                stderr=subprocess.PIPE,
+                start_new_session=True
             )
             server_processes.append(process)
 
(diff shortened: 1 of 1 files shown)


98zc5g5jyw-arch:
## Babysit review — LGTM ✅

`start_new_session` + `os.killpg` (with SIGKILL fallback and win32 branch) properly kills the server's process group — real fix for orphaned servers, more complete than the overlap in #1602. Good.

98zc5g5jyw-arch (approved):
## Hermes Review

**Verdict: Approve** — correct and well-handled fix for orphaned server processes.

### Verified
- `start_new_session=True` puts the shell + children in a fresh process group; `os.killpg(pgid, SIGTERM)` then reaches the real server, not just the shell — fixes the orphaned-port problem described.
- `ProcessLookupError` is caught both at `getpgid` time and around `killpg`/`terminate()` — the fallback path is safe for already-dead processes.
- SIGTERM → 5s wait → SIGKILL escalation is a sane cleanup ladder.

- Consider also `os.setpgid`-independent portability note: `start_new_session` is POSIX-only; on Windows this file's behavior changes (killpg unavailable) — a short comment or guard would help Windows contributors.
- `process.terminate()` in the `pgid is None` branch can still raise `ProcessLookupError`, but it's inside the `try` → covered by the `except (TimeoutExpired, ProcessLookupError)` handler. Good.

PR #1598, opened 2026-08-17
plan-file-hygiene: add post-edit verification (follow-up to #1479)

halilxibrahim:
Follow-up to #1479 (plan-file-hygiene skill, addresses #1417) — picks up the thread where @tonydzi and I discussed adding a check that runs after an edit lands, not just before it.

Rule 2 guards against clobbering before a write. This adds Rule 5: Verify after you write — diff the file immediately before/after an edit to confirm only the intended change is present and nothing else moved.
Adds one line to "Report the hygiene pass" so the post-write diff check is reported too.

Stacked on [#1479](https://github.com/halilxibrahim/skills/issues/1479)'s commit since plan-file-hygiene doesn't exist on main yet — only the last commit is new. Happy to rebase once #1479 merges, or fold into #1479 directly if you prefer, @tonydzi.

### DIFF
--- a/skills/plan-file-hygiene/LICENSE.txt
+++ b/skills/plan-file-hygiene/LICENSE.txt
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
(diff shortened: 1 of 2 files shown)


98zc5g5jyw-arch:
## Babysit review — LGTM ✅

Post-edit verification step is well specified (re-read + syntax check), LICENSE included (Apache-2.0), description within limits. Follow-up to #1479 is coherent.

98zc5g5jyw-arch (approved):
Reviewed by Hermes Agent. Plan-file-hygiene follow-up: Rule 5 (post-edit verification) is well-specified and correctly closes the clobbering gap from #1479. Frontmatter valid (name/description/license present, description within limits), LICENSE.txt Apache-2.0. Good quality doc. LGTM.

Issue #1467, opened 2026-07-23
skill-creator: quick_validate.py imports PyYAML but it's not declared as a dependency anywhere


100kimch:
`scripts/quick_validate.py` does:

```python
import yaml
...
frontmatter = yaml.safe_load(frontmatter_text)
...
except yaml.YAMLError as e:
```

`yaml` (PyYAML) is not part of the Python standard library, and neither `skill-creator/` nor the repo root declares it as a dependency anywhere (no `requirements.txt`, `pyproject.toml`, etc. in this repo). On a machine that doesn't happen to have PyYAML already installed globally, running the skill's validation step fails with `ModuleNotFoundError: No module named 'yaml'`.

We hit this while vendoring `skill-creator` into an internal docs repo and auditing which external packages our agent skills actually require — this was the only "silently required but undeclared" dependency we found.

Suggested fix: add a `requirements.txt` (or equivalent) alongside `skill-creator/scripts/quick_validate.py` pinning `pyyaml`, so anyone installing/running this skill standalone (e.g. via `gh skill install`) knows to install it first.

Related but distinct from #150 (which is about whether *skill frontmatter itself* should support a `dependencies` field) — this issue is specifically about the validator script's own missing dependency declaration.

PR #1599, opened 2026-08-17
skill-creator: declare pyyaml dependency for quick_validate.py

halilxibrahim:
## Summary
- `skill-creator/scripts/quick_validate.py` imports `yaml` (PyYAML), which is not part of the standard library and wasn't declared as a dependency anywhere in the repo.
- Adds `skills/skill-creator/scripts/requirements.txt` pinning `pyyaml>=6.0`, matching the existing pattern used by `mcp-builder/scripts/requirements.txt`.

Fixes #1467

## Test plan
- [x] Confirmed `quick_validate.py` is the only script under `skill-creator/` that imports `yaml`
- [x] Verified the `requirements.txt` format matches the convention already used elsewhere in the repo (e.g. `mcp-builder/scripts/requirements.txt`)


98zc5g5jyw-arch:
## Babysit review — LGTM ✅

Declaring `pyyaml>=6.0` is the right fix for quick_validate.py. Note this is one of three equivalent PRs for the same issue (#1599/#1602/#1472) — merging any one resolves it; #1602's hand-rolled fallback parser should be dropped in favor of this dependency.

98zc5g5jyw-arch (approved):
## Hermes Review

**Verdict: Approve** — dependency correctly declared.

### Verified
- `quick_validate.py` imports `yaml`; the new `requirements.txt` with `pyyaml>=6.0` covers it. Correct and minimal.

### Suggestion
- Duplication: #1556, #1472 and this PR all declare the PyYAML dependency in different forms (requirements.txt vs. inline doc). Coordinate so maintainers merge one canonical form (requirements.txt is the cleanest).

Issue #1475, opened 2026-07-24
claude-api: prompt-caching guidance conflicts with published docs on priming the cache for Message Batches


claudeopusagora:
The `claude-api` skill's `shared/prompt-caching.md` recommends priming/pre-warming the cache with a **regular (non-batch) Messages request** (the "Concurrent-request timing" fan-out pattern and the `max_tokens: 0` pre-warming section), with no caveat about the Batch API. The published docs recommend a different transport for batch workloads: prime via a **single-request batch with a 1-hour cache block**, then submit the rest ([prompt caching → cache storage and sharing](https://platform.claude.com/docs/en/build-with-claude/prompt-caching#cache-storage-and-sharing)).

We measured the difference today (claude-haiku-4-5, first-party API):

- A regular Messages-API request wrote an ~11k-token shared prefix (tools+system, 1h breakpoint); a batch submitted **1 second later** with a byte-identical prefix read **0** cached tokens and re-wrote all 11,189 (`cache_read_input_tokens: 0`, `cache_creation.ephemeral_1h_input_tokens: 11189`).
- Entries written **by batch items** were reliably read by later batch submissions (11,189–19,933 token reads across successive submissions, minutes apart).

Anecdotally, cross-surface priming (Messages write → batch read) *used* to work — a previous version of our runner relied on it — so this looks like an implementation change the skill hasn't caught up with. Suggested fix: add a note to the pre-warming / concurrent-timing sections that for Batch API workloads the prime should itself be a (single-request) batch, matching the published docs, and that Messages↔Batch cache sharing should not be relied on.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Lerv4Wj3Gy54UGzkfpRVVc

PR #1601, opened 2026-08-17
claude-api: note Batch API cache-priming caveat in prompt-caching.md

halilxibrahim:
## Summary
- `skills/claude-api/shared/prompt-caching.md`'s "Concurrent-request timing" and "Pre-warming the cache" sections recommended priming the cache with a regular (non-batch) Messages request, with no caveat about the Batch API.
- Per the published docs ([prompt caching → cache storage and sharing](https://platform.claude.com/docs/en/build-with-claude/prompt-caching#cache-storage-and-sharing)), batch workloads should be primed with a single-request batch instead — Messages↔Batch cache sharing should not be relied on.
- Verified by measurement: a regular Messages request wrote an ~11k-token shared prefix, but a batch submitted 1 second later with a byte-identical prefix read 0 cached tokens (rewrote the entire prefix). Entries written by batch items, however, were reliably read by later batch submissions.
- Added a note to both sections covering this caveat; no code/behavior change, documentation only.

Fixes #1475

### DIFF
--- a/skills/claude-api/shared/prompt-caching.md
+++ b/skills/claude-api/shared/prompt-caching.md
@@ -200,6 +200,8 @@ A cache entry becomes readable only after the first response **begins streaming*
 
 For fan-out patterns: send 1 request, await the first streamed token (not the full response), then fire the remaining N−1. They'll read the cache the first one just wrote.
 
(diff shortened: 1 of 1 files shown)


98zc5g5jyw-arch:
## Babysit review — LGTM ✅

Batch API cache-priming caveat matches the official docs (cache entries aren't reliably shared Messages↔Batch); the suggestion to prime with a single-request batch is correct. Clean 4-line doc addition.

98zc5g5jyw-arch (approved):
## Hermes Review

**Verdict: Approve** — accurate, well-cited documentation addition.

### Verified
- The Batch API cache caveat matches the documented cache-storage-and-sharing semantics (Messages requests and Message Batches don't reliably share cache entries). The pre-warm note (`max_tokens: 0` is itself a Messages request) is a correct and important consequence.
- Cites the official docs page inline — good practice.

PR #1609, opened 2026-08-18
skill-creator: trigger eval mounts the probe as a skill, not a command

headinthebox:
## Problem

`run_eval.py` writes the description under test into `.claude/commands/`. Commands are **user-invoked slash commands**: on current Claude Code versions (measured on 2.1.234, macOS) they never surface in the model's `available_skills`, so no query ever triggers and every description scores identically. Measured symptom: a full `run_loop.py` run over 20 queries × 5 iterations kept train and test scores frozen at exactly the count of should-not-trigger queries — for the original description, for hand-written pushier ones, and for the loop's own increasingly aggressive proposals. Every should-trigger query sat at `trigger_rate: 0.0` throughout. A score that cannot move is measuring the mount, not the skill, and `best_description` degrades to a tie-break on the original.

## Fix

Mount a transient `SKILL.md` under `.claude/skills/<name>-<uuid6>/` instead. The existing stream detection already watches the `Skill` tool, and the catalog keys on the directory name, so the rest of the pipeline is unchanged.

Two secondary findings folded in:

- **The unique suffix stays short and word-free.** The catalog entry is the directory name; in probes, a long machine-looking mount (`analysis-skill-deadbeef`) went unconsulted — the model inspected the file with `cat` rather than invoke it — while the same description under a plain or short-suffixed name (`analysis-zz9x4q`) triggered first-thing. (Confounded with description strength across probes, so recorded as naming guidance in a comment rather than claimed as a hard result.)
- **The project root must be a folder the user has previously trusted in Claude Code.** Project skills are not loaded from untrusted folders, which silently reproduces the same all-zero symptom. Documented in `find_project_root`'s docstring.

## Verification

End to end on a real skill (a project-internal analysis guide), serial and with parallel workers:

| | before (commands mount) | after (skills mount) |
|---|---|---|
| should-trigger query | 0.0 — for every description tried | 1.0 |
| near-miss negative | 0.0 (vacuously) | 0.0 (genuinely — the entry is visible and not consulted) |

With the fix in place the eval discriminates descriptions: a deliberately thin description measured 0/10 on realistic should-trigger queries, a rewritten one 4/10 with all 10 near-miss negatives held — the before/after the tool exists to produce.

### DIFF
--- a/skills/skill-creator/scripts/run_eval.py
+++ b/skills/skill-creator/scripts/run_eval.py
@@ -9,6 +9,7 @@ import argparse
 import json
 import os
 import select
+import shutil
 import subprocess
 import sys
 import time
@@ -22,8 +23,11 @@ from scripts.utils import parse_skill_md
 def find_project_root() -> Path:
     """Find the project root by walking up from cwd looking for .claude/.
 
-    Mimics how Claude Code discovers its project root, so the command file
-    we create ends up where claude -p will look for it.
+    Mimics how Claude Code discovers its project root, so the skill we mount
+    ends up where claude -p will look for it. Note: the project root must be
+    a folder the user has previously TRUSTED in Claude Code — project skills
+    are not loaded from untrusted folders, and the eval then reports a zero
+    trigger rate for every query regardless of the description under test.
     """
     current = Path.cwd()
     for parent in [current, *current.parents]:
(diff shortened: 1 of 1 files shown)


98zc5g5jyw-arch:
## Babysit review — mechanism conflict + no tests

Mounting the probe as a *skill* instead of a *command* is mutually exclusive with #1630/#1629's command-probe approach (same run_eval.py territory), and it doesn't cover the "skill already installed" scenario (the 0% recall bug). SKILL.md still documents the old `commands` usage while the code switched to skill mounting — docs not synced. Zero tests. Suggest: either rebase onto #1630 and demonstrate an actual failing eval before/after, or close in favor of #1630.

98zc5g5jyw-arch (commented):
## Hermes Review

**Verdict: Comment** — well-reasoned fix; coordinate with the parallel eval-harness PRs.

### Verified
- `.claude/skills/<name>/SKILL.md` is the correct location for Claude Code project skills, and `.claude/commands/` are user-invoked slash commands that do not surface in `available_skills` — the 0% trigger-rate diagnosis matches observed Claude Code behavior (commands ≠ skills in the model's tool list).
- Per-run unique mount dir (`-{uid6}`) prevents parallel-worker filesystem collisions and makes cleanup safe (`shutil.rmtree` on own dir only).
- Frontmatter YAML block-scalar escaping is preserved correctly.

### Warnings
- **Heavy overlap with #1630 and #1629**, which rewrite the same `run_eval.py` regions (both touch 10+ of the same files). These three PRs will conflict; please coordinate the canonical version of the mount logic before all three pile onto `main`.

- `skill_name` comes from CLI input — if it contains characters illegal in a skill folder name (spaces, `/`), the mount path breaks. A tiny sanitizer would harden it.
- The docstring now says "must be a folder the user has previously TRUSTED" — consider also surfacing that as a runtime error instead of only a docstring, since it silently yields 0% trigger rates otherwise (that's exactly the trap #1629 describes).

PR #1610, opened 2026-08-19
Render Markdown and Mermaid in the skill-creator eval viewer

zengj8:
## Summary
- Eval viewer previews Markdown outputs instead of dumping them as raw text.
- Fenced mermaid blocks render as diagrams on the Outputs tab.
- `.md` files are classified as markdown so the viewer can pick that renderer.

## Test plan
- [ ] Generate a review page from a workspace whose outputs include a `.md` file with a fenced mermaid block.
- [ ] Confirm the diagram renders on the Outputs tab.
- [ ] Confirm a mermaid parse error leaves the source visible.
- [ ] Confirm non-markdown text files still render as `<pre>`.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/skills/skill-creator/SKILL.md
+++ b/skills/skill-creator/SKILL.md
@@ -254,7 +254,7 @@ Note: please use generate_review.py to create the viewer; there's no need to wri
 
 The "Outputs" tab shows one test case at a time:
 - **Prompt**: the task that was given
-- **Output**: the files the skill produced, rendered inline where possible
+- **Output**: the files the skill produced, rendered inline where possible. Markdown is previewed; fenced mermaid blocks render as diagrams.
 - **Previous Output** (iteration 2+): collapsed section showing last iteration's output
 - **Formal Grades** (if grading was run): collapsed section showing assertion pass/fail
 - **Feedback**: a textbox that auto-saves as they type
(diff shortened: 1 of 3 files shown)


98zc5g5jyw-arch:
## Babysit review — XSS surface, needs changes

marked@9.1.6 is rendered via `innerHTML` without sanitization, and mermaid is configured with `securityLevel: "loose"` (HTML injection possible) — for an eval viewer that renders arbitrary skill content this is a real XSS surface. Also: CDN scripts have no SRI, and the change removes the `pre-wrap` fallback styling. This also conflicts directionally with #1589's escapeHtml hardening. Recommend: DOMPurify before innerHTML, mermaid `securityLevel: "strict"`, add SRI or vendor the libs.

98zc5g5jyw-arch (commented):
## Hermes Review

**Verdict: Comment** — good feature; two hardening nits on the CDN additions.

### Verified
- `embed_file` tagging `.md` outputs as `markdown` type and viewer.html rendering them via `marked` is a clean, minimal change; the mermaid renderer path is wired consistently with the existing viewer structure.
- The existing xlsx script in the same file uses an **SRI `integrity` + `crossorigin`** attribute; the two new CDN scripts (`marked@9.1.6`, `mermaid@10.9.3`) do **not**. For consistency and supply-chain safety, add SRI hashes.

### Warnings
- `mermaid@10.9.3` is a dated major (current line is v11). Pin the latest v11 or the specific version you validated; also verify both libs are needed on the eval-viewer page load vs. lazy-loaded only when a mermaid block exists.

- If the viewer is used in offline/sandboxed eval environments, CDN-only loading will silently degrade the new preview feature — consider a graceful fallback note (e.g., "if CDN unavailable, show raw text").

PR #1612, opened 2026-08-19
Add lnurl-auth skill (LUD-04 authentication)

dyegolara:
Educational skill that explains and guides an auth-only LNURL-auth (LUD-04) login — what it is, the handshake (decode bech32, get k1, sign with secp256k1, callback GET), and its limits (no wallet, no node, no payment; not a payment flow). Markdown-only, no scripts or external dependencies.

### DIFF
--- a/skills/lnurl-auth/SKILL.md
+++ b/skills/lnurl-auth/SKILL.md
@@ -1,0 +1,47 @@
+---
+name: lnurl-auth
+description: "Explain and guide an auth-only LNURL-auth (LUD-04) login when a site provides a Sign in with Lightning challenge."
+---
+
+# LNURL-auth (LUD-04)
+
+Use this skill when a service presents a `lnurl1...` value for Sign in with
+Lightning. LNURL-auth is an authentication handshake, not a payment flow: it
+does not create an invoice, spend funds, require a wallet, or require a
+Lightning node.
+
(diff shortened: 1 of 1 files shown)


98zc5g5jyw-arch:
## Babysit review — LGTM ✅

LUD-04 details are correct (k1 not re-hashed, DER signature, one-time k1), security boundaries explicit (0600 perms, per-domain derived keys, no payment flow). Low risk.

98zc5g5jyw-arch (commented):
## Hermes Review

**Verdict: Comment** — solid protocol skill; one spec-compliance nit.

### Verified
- LUD-04 flow is accurate: bech32 decode → 32-byte `k1` (sign raw, no re-hash) → secp256k1 signing key → DER-encoded sig + compressed pubkey → callback GET → `{"status":"OK"}`.
- Safety guidance is good (per-domain derived keys to prevent correlation, no invoice payments, one-time `k1`).

### Warnings
- **Missing `license:` frontmatter key and `LICENSE.txt`.** Repo convention is `name`/`description`/`license` (+ LICENSE.txt). Add both.

- Consider noting explicitly that `key=` must be the **compressed** public key hex (it's stated) and that some services expect the callback via HTTP GET only (already implied). Fine as-is otherwise.

Issue #1613, opened 2026-08-19
claude-api: frontmatter description exceeds 1024-character limit


ilvpc2004:
## Summary

`skills/claude-api/SKILL.md` currently has a 1,068-character `description` value in its YAML frontmatter. The Agent Skills specification limits `description` to 1,024 characters.

Specification: https://agentskills.io/specification#description-field

## Reproduction

1. Fetch the current `skills/claude-api/SKILL.md` from this repository.
2. Parse the YAML frontmatter.
3. Measure `len(description)`.
4. Actual: `1068`; allowed maximum: `1024`.

## Impact

Strict Agent Skills validators and clients may reject or skip the skill metadata.

## Suggested fix

Shorten the frontmatter description to at most 1,024 characters while preserving the key positive and negative trigger conditions. The skill body does not need to change.

BRGOVIND:
Hi! I’d like to work on this issue. I’ll shorten the `claude-api` frontmatter description to comply with the 1,024-character limit while preserving the existing trigger conditions, without changing the skill body.

PR #1596, opened 2026-08-17
fix(skill-creator): enforce utf-8 encoding across scripts and fix claude-api max description length

aoright:
### Summary
1. **Explicit UTF-8 Encoding**: Add `encoding="utf-8"` to all `.read_text()` and `.write_text()` calls in `skill-creator` scripts (`quick_validate.py`, `utils.py`, `run_eval.py`, `improve_description.py`, `run_loop.py`, `generate_report.py`) to prevent cross-platform `UnicodeDecodeError` on Windows systems.
2. **Specification Compliance**: Condense `skills/claude-api/SKILL.md` description to fit within the 1024 character limit specified by the skill validation schema (reduced from 1068 to 870 chars) while preserving all triggers, skips, and instructions.

### Validation
- Validated all 17 skills in the repository via `quick_validate.py` (17/17 passed).

PR #1616, opened 2026-08-20
Fix/claude api description limit

BRGOVIND:
## Summary

* Shorten the `claude-api` skill frontmatter description to comply with the 1,024-character limit.
* Preserve the existing positive and negative trigger conditions.
* Leave the skill body unchanged.

## Validation

* `python -X utf8 skills\skill-creator\scripts\quick_validate.py skills\claude-api` → `Skill is valid!`
* `git diff --check` passes.

Fixes #1613

### DIFF
--- a/skills/claude-api/SKILL.md
+++ b/skills/claude-api/SKILL.md
@@ -2,7 +2,7 @@
 name: claude-api
 description: |-
   Reference for the Claude API / Anthropic SDK — model ids, pricing, params, streaming, tool use, MCP, agents, caching, token counting, model migration.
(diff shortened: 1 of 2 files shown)


98zc5g5jyw-arch:
## Babysit review — overlap with #1596, plus scope note

This trims the claude-api description to 1013 chars (verified < 1024 ✅). However #1596 rewrites the SAME description line (876 chars) plus adds skill-creator utf-8 fixes, so the two will conflict on the identical line — recommend merging #1596 and closing this one, or rebasing this on top keeping only what #1596 doesn't cover. Also note the PPTX COM-automation warning is a separate concern bundled into this PR (fine content, but splits review focus).

98zc5g5jyw-arch (commented):
## Hermes Review

**Verdict: Comment** — description trim verified OK, but this overlaps with the same trim in #1635 / #1596; the pptx addition is unrelated but good.

### Verified
- New description measures **537 chars** (main is 1074 > 1024 cap) — fits.
- The pptx COM-automation warning (never drive a user's open Office instance; install LibreOffice instead) is accurate and valuable — `Quit()` on a shared instance is a known data-loss footgun.

### Warnings
- The TRIGGER trim removes "don't skip because it 'looks like a one-liner'" — a deliberate anti-skip emphasis. If you want to keep it, shorten the SKIP clause instead.

### Suggestions
- Duplication cluster: #1635 (839 chars), #1596 (876 chars) trim the same description. Coordinate so only one lands.
- Suggest splitting the pptx change into its own PR — mixing an unrelated safety doc into a description-trim PR slows review.

PR #1623, opened 2026-08-21
Update claude-api skill: Python SDK 0.x to 1.x upgrade guide

cj-ant:
## What this does

Adds an `upgrade` subcommand to the `claude-api` skill for moving a Python project from `anthropic` 0.x to 1.x, and updates the Python README for the 1.x HTTP layer.

## Why now

`anthropic` 1.0.0 is on PyPI. It moves the HTTP layer from `httpx` to its maintained fork `httpx2`, raises the floor to Python 3.10, and removes long-deprecated surface (Text Completions, `LegacyAPIResponse`, several aliases and parameters). Most edits are mechanical, but an `httpx.Timeout` or `httpx.Client` handed to a 1.x client fails at request time rather than import time, and HTTP mocking or tracing that patches `httpx` silently stops seeing SDK traffic. The skill's `migrate` subcommand covers model migrations, not package upgrades, so there was nothing to route these requests to.

## What changed

**New file: `python/claude-api/sdk-upgrade.md` (286 lines).**

- Step 0 confirms scope (same ask-first rule as `migrate`), reads the current version, and checks that a 1.x release is published before writing a pin.
- Step 1 is an inventory table: one row per signal to grep for (`import httpx`, `with_raw_response`, `completions.create`, `respx` / `pytest_httpx` / OpenTelemetry `httpx` instrumentation, and so on), each pointing at the section that handles it. Re-run at the end, it doubles as verification.
- Steps 2 to 8 cover the dependency bump, `httpx` to `httpx2` objects, awaited async `.with_raw_response`, Text Completions removal, removed sampling parameters (with the `extra_body` escape hatch for callers pinned to an older model), renamed exports, and smaller call-site changes.
- Closing sections cover verification and the report format. The SDK repo's `MIGRATION.md` stays authoritative and the guide says to defer to it on any disagreement.

**`SKILL.md` (2 edits).**

- New `upgrade` row in the subcommand table. Trailing words can name a language or scope (`upgrade python src/`). A language with no bundled guide gets pointed at its SDK CHANGELOG rather than an improvised port of the Python steps. The row states this is not model migration.
- New reading-guide entry for "upgrading the SDK package across a major version, or writing new code against a project already on 1.x".

**`python/claude-api/README.md` (3 edits).** Timeouts use `anthropic.Timeout` instead of `httpx.Timeout` (and the stray `import httpx` is gone), the async and custom-client notes name `httpx2`, and a short paragraph explains that `anthropic.Timeout` is `httpx2.Timeout` with a link to `MIGRATION.md` and `/claude-api upgrade python`.

**`shared/live-sources.md` (1 edit).** New "SDK major-version upgrade guides" table with the Python `MIGRATION.md` URL and extraction prompt.

The frontmatter `description` is unchanged. No unresolved template placeholders.


rlancemartin (approved):

wissamblue69-dotcom (approved):

skills/claude-api/python/claude-api/README.md
```diff
@@ -41,24 +41,24 @@ client.with_options(timeout=5.0, max_retries=5).messages.create(
 
 ### Timeouts
 
-Default request timeout is 10 minutes. Pass a float (seconds) or an `httpx.Timeout` for granular control. On timeout the SDK raises `anthropic.APITimeoutError` (and retries per `max_retries`).
```
Alsami1998: Twww

PR #1628, opened 2026-08-21
Add Hivemind: Zero-Cost Multi-Agent Orchestration Skill

Hanishchow:
Adds **Hivemind**, a skill that lets Claude Code delegate mechanical work to headless [opencode](https://opencode.ai) workers running on free models, while Claude Code stays the only planner, reviewer, and merger.

## Why

The expensive model's context is the scarce resource, not its intelligence. Hivemind moves grunt work — multi-lens review, research sweeps, bulk migrations, partitioned test runs — onto disposable workers, and every worker returns **exactly one compact JSON line**. Raw agent streams never enter the orchestrator's context, which is the whole point of the protocol.

## What's in it

```
skills/hivemind/
  SKILL.md              invocation contract, swarm flow, fallback ladder, anti-patterns
  LICENSE.txt           Apache 2.0
  scripts/oc-worker.mjs the single sanctioned worker entry point
  scripts/oc-status.mjs fleet progress recovered from run logs
  scripts/oc-aggregate.mjs   consensus-first synthesis across N workers
  scripts/bench/        A/B/C benchmark harness + blind grading rubric
  assets/commands/      the 7 slash-command entry points
  assets/agents/        scout / coder / tester worker personas
```

Three personas, each least-privilege: **scout** (read-only, no write/edit/bash), **coder** (writes, confined to one git worktree), **tester** (runs tests, never edits source). Writing workers never share a directory, and the orchestrator is the only merger.

## Dependency disclosure

This is an orchestration layer over the third-party `opencode` CLI, and worker traffic goes to opencode's endpoints. SKILL.md opens with a Prerequisites section stating that plainly, including that the free tier belongs to opencode rather than Anthropic and that availability, rate limits, and pricing are theirs to change. It also warns against delegating secrets or private code. Flagging it here too rather than leaving it to be discovered in review.

## Canonical home

Maintained at **https://github.com/Hanishchow/hivemind** (Apache 2.0). This PR is a snapshot; fixes land there first and get ported here.

Contributor: @Hanishchow

### DIFF
--- a/skills/hivemind/.gitignore
+++ b/skills/hivemind/.gitignore
@@ -1,0 +1,2 @@
+.runs/
+bench-results.jsonl
--- a/skills/hivemind/LICENSE.txt
+++ b/skills/hivemind/LICENSE.txt
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
+
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
+
+   2. Grant of Copyright License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      copyright license to reproduce, prepare Derivative Works of,
+      publicly display, publicly perform, sublicense, and distribute the
+      Work and such Derivative Works in Source or Object form.
+
+   3. Grant of Patent License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      (except as stated in this section) patent license to make, have made,
+      use, offer to sell, sell, import, and otherwise transfer the Work,
+      where such license applies only to those patent claims licensable
+      by such Contributor that are necessarily infringed by their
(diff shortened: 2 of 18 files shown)


98zc5g5jyw-arch:
## Babysit review — needs path fix

Nice content, but the new files live at the repo **root** `hivemind/` instead of `skills/hivemind/` — the repo's skill scanner won't pick them up, so the skill is effectively invisible after merge. Also: `.runs/hive-smoke.jsonl` is a runtime artifact and shouldn't be committed; and `oc-worker.mjs` defaults to an external opencode model (`mimo-v2.5-free`) — please document that dependency. Move the directory, drop the `.runs/` trace, and re-push.

98zc5g5jyw-arch (changes requested):
## Review: Hivemind skill (#1628)

Thanks for the submission — the skill is well-structured, the scripts are defensive (argv-array spawning, `--dir` validation, timeouts, capped stderr, NDJSON parsing with fallbacks), and the "Golden Rule / fallback ladder / worktree isolation" protocol is genuinely good engineering. No credentials or secrets are hardcoded anywhere. Two blocking items below (both cheap to fix), then non-blocking notes.

### Blocking

1. **Missing `license` key in `SKILL.md` frontmatter** — the repo convention (see `skills/claude-api/SKILL.md` and `skills/docx/SKILL.md`) requires three keys:
   ```yaml
   ---
   name: hivemind
   description: ...
   license: Complete terms in LICENSE.txt
   ---
   ```
   The current frontmatter has only `name` and `description`. Please add the `license` key.

2. **Missing `LICENSE.txt`** — every skill directory in this repo ships a `LICENSE.txt` with complete terms (`skill-creator`, `claude-api`, `docx`, ...). The PR's 17 files contain no license file. Please add `skills/hivemind/LICENSE.txt` with the appropriate terms for the skill content.

### Non-blocking

3. **`description` length/format** — 390 chars, single line, valid YAML (under the ~1024-char limit, so it passes), but it is on the long side; consider tightening the first sentence to the core trigger ("Orchestrate free opencode workers from Claude Code…") and moving the longer trigger phrases later, matching the repo's "short single-paragraph description" convention. Not a blocker.

4. **Security hardening (suggestion)** — in `scripts/oc-worker.mjs`, `ensureServer()` runs `spawn(\`${quoted} serve --port ${PORT}\`, { shell: true, ... })`. `PORT` is derived from the `HIVEMIND_SERVER_URL` env var, so a hostile value for that env var could inject shell syntax. Low severity (local env only), but consider spawning with an args array and `shell: false`, or validating that `PORT` is numeric. The main `opencode run` invocation already uses an args array (no shell interpolation) — good.

5. **Third-party dependency note** — this skill is an orchestration layer over the third-party `opencode` CLI and routes worker traffic to opencode's endpoints/free models. The SKILL.md discloses this clearly ("Do not delegate secrets or private code you would not send there") and caveats that free-tier pricing/availability is opencode's to change — thank you for that. Just flagging for maintainer awareness; acceptable as disclosed.

6. **No duplication** — checked the existing skills in the repo; no agent-collaboration/orchestration skill exists, so this fills a genuine gap. Directory layout (`assets/agents/`, `assets/commands/`, `scripts/`, `scripts/bench/`) is consistent with other script-heavy skills.

### Summary

Great first contribution. Fix items 1 and 2 (add `license` key + `LICENSE.txt`), and this should be good to merge.

Hanishchow:
Thanks for the review — all three points are addressed, force-pushed as a single verified commit rebased on current `main`.

- **Path:** `hivemind/` → `skills/hivemind/`, so the skill scanner picks it up.
- **Runtime artifact:** `.runs/hive-smoke.jsonl` removed, and `.runs/` is now gitignored inside the skill folder so traces can't be committed again.
- **External dependency:** SKILL.md opens with a Prerequisites table covering Node >= 18, the `opencode` CLI, authentication, and the default `opencode/mimo-v2.5-free` model — including that the free tier belongs to opencode (not Anthropic) and that availability, rate limits, and pricing are theirs to change. It also states plainly that worker traffic leaves for opencode's endpoints, so secrets and private code shouldn't be delegated.

Two things I fixed while in there, since they'd have been the next round of review:

- The skill referenced seven slash commands and three worker agents that **weren't in the PR at all**. They now ship under `assets/commands/` and `assets/agents/`, with a Setup section explaining where to copy them. Their invocations previously carried an absolute `C:\Users\<me>\...` path; that's now `$HIVEMIND_HOME`.
- `resolveOpencode()` probed with `where.exe` unconditionally, so it silently fell back to a bare `opencode` on macOS and Linux. It now uses `which` off-Windows and keeps the `.cmd`-shim parsing on Windows (Node's EINVAL policy blocks spawning `.cmd` directly).

Hanishchow:
Both blocking items are fixed, plus 3 and 4. Pushed as a follow-up verified commit.

**1. `license` frontmatter key** — added:

```yaml
name: hivemind
description: ...
license: Complete terms in LICENSE.txt
```

**2. `LICENSE.txt`** — added at `skills/hivemind/LICENSE.txt`, Apache 2.0, byte-identical to the copy in `skills/canvas-design` apart from the copyright line, which reads `Copyright 2026 N Hanish (@Hanishchow).` rather than Anthropic PBC, since I'm contributing the content rather than assigning it. If you'd rather it match the other skills exactly, say so and I'll change it in one push — no objection either way.

**3. `description`** — 390 → 300 chars, core trigger first:

> Orchestrate free opencode workers from Claude Code to cut token costs. Use when delegating grunt work to a single worker or a parallel swarm (scout/coder/tester) with worktree isolation, benchmarking against opencode, or when the user says "spawn a worker", "swarm", "delegate to opencode", or "/oc".

**4. Shell injection** — you were right to flag it, and chasing it turned up a second, more likely bug. `ensureServer()` now spawns with an args array and `shell: false`, so nothing is interpolated:

```js
const child = spawn(bin, ["serve", "--port", PORT], { shell: false, detached: true, stdio: "ignore", windowsHide: true });
```

That also removed the manual path-quoting the shell form needed. `PORT` is now produced by a `resolvePort()` that rejects anything non-numeric or out of range.

The second bug: a malformed `HIVEMIND_SERVER_URL` made `new URL()` throw at module top level, so the script died with a stack trace on stdout/stderr — breaking the one-JSON-line contract the whole skill depends on. Now:

```
$ HIVEMIND_SERVER_URL="not a url" node scripts/oc-worker.mjs "task"
{"ok":false,"stage":"args","error":"HIVEMIND_SERVER_URL is not a valid URL","result":"",...}

$ HIVEMIND_SERVER_URL="http://127.0.0.1:70000" node scripts/oc-worker.mjs "task"
{"ok":false,"stage":"args","error":"HIVEMIND_SERVER_URL is not a valid URL","result":"",...}
```

The env var and its validation are now documented in SKILL.md next to the server-management paragraph.

On **5** — noted, and I'd rather it stay flagged than buried. If the third-party dependency ever becomes disqualifying for this repo, tell me and I'll withdraw it without argument.

98zc5g5jyw-arch (approved):
Re-review after b4f0c6f7: all previously requested changes are addressed. 1) license key added to frontmatter, 2) LICENSE.txt shipped, 3) description tightened to 373 chars, 4) oc-worker.mjs now validates PORT via resolvePort() and spawns with shell:false args array. Looks good. LGTM.

PR #1629, opened 2026-08-22
skill-creator: fix three eval-harness defects that silently report a 0.0 trigger rate

nic21vdw:
Three defects in `skill-creator`'s eval harness that share a failure mode: the run completes, prints well-formed JSON, and reports numbers that are wrong. Nothing errors out, so the output reads as a verdict on the skill rather than a broken measurement.

I hit these trying to measure trigger accuracy for seven skills I already had installed. The first run reported a 0.0 trigger rate across the board, which looked like a real result.

---

## 1. `--use-installed`: the eval cannot measure a skill that is installed

`run_single_query` writes a uniquely-named copy of the skill into `.claude/commands/` and checks whether Claude invoked *that* name:

```python
clean_name = f"{skill_name}-skill-{unique_id}"
...
if tool_name == "Skill" and clean_name in tool_input.get("skill", ""):
```

That holds only while the skill under test is **not** on the load path. When it is already installed, Claude invokes the real skill. Captured from `claude -p --output-format stream-json --include-partial-messages`, evaluating a `git-rescue` skill installed at user scope:

```json
{"type":"content_block_start","content_block":{"type":"tool_use","name":"Skill","input":{}}}
{"type":"content_block_delta","delta":{"type":"input_json_delta","partial_json":"{\"skill\": \"git-rescue"}}
{"type":"content_block_delta","delta":{"type":"input_json_delta","partial_json":"\", \"args\": \"three"}}
{"type":"content_block_delta","delta":{"type":"input_json_delta","partial_json":" commits landed on main in CoLateral repo, need to move them onto a branch"}}
{"type":"content_block_stop"}
```

The skill triggered, correctly and on the first turn. But the accumulated input is `{"skill": "git-rescue", ...}` and `clean_name` is `git-rescue-skill-4f2a91c8`, so the substring check fails and the query is scored as a miss. Every query fails identically, and the eval reports `"trigger_rate": 0.0` for a skill that fires exactly as intended.

This is the state any user is in when auditing skills they already use — which is what the description-optimization workflow in `SKILL.md` is for.

`--use-installed` matches the skill by its real name and skips writing the probe copy. On the same query and skill, the trigger rate goes from `0.00` to `1.00`.

It is rejected in combination with `--description`: a candidate description reaches Claude *through* the probe copy, so with `--use-installed` the installed skill supplies the description and the override would be silently ignored. `run_loop.py` is unchanged and still uses the probe, since testing candidate descriptions depends on it. `SKILL.md` documents both the flag and that constraint.

## 2. `select()` on a pipe is not portable

`run_single_query` polls the subprocess pipe with `select.select([process.stdout], [], [], 1.0)`. On Windows `select` accepts sockets only, so the first query raises:

```
OSError: [WinError 10093] Either the application has not called WSAStartup, or WSAStartup failed
```

`run_eval` catches this per future, warns on stderr, and appends `False`:

```python
except Exception as e:
    print(f"Warning: query failed: {e}", file=sys.stderr)
    query_triggers[query].append(False)
```

So the harness produces a complete, well-formed result in which every trigger rate is 0.0 — indistinguishable from a description that never fires. Replaced with a reader thread feeding a queue, preserving the existing timeout behaviour.

## 3. File IO relies on the platform default encoding

29 `read_text()` / `write_text()` / `open()` calls across `scripts/` and `eval-viewer/` do not name an encoding. On Windows that resolves to cp1252, so any skill containing a character outside it fails to read:

```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 4633:
character maps to <undefined>
```

Several skills in this repository use em dashes and curly quotes in `SKILL.md`, so on Windows `quick_validate.py` cannot read them — including skills shipped in the same `example-skills` bundle as `skill-creator` itself.

---

## Structure

Three commits, one per defect. The `--use-installed` commit also carries the tests, since the refactor it needs is what makes detection testable.

To keep the fix honest rather than incidental, `run_single_query` now delegates to three extracted pieces — `TriggerDetector` / `detect_trigger()` for the stream state machine, `iter_process_lines()` for the pipe, and `resolve_probe_name()` for the name choice. The state machine logic is unchanged; it is only lifted out so it can be run against a recorded transcript instead of a live subprocess.

## Tests

New `skills/skill-creator/tests/`, standard library only, no new dependencies:

```bash
cd skills/skill-creator && python -m unittest discover -s tests
```

12 tests, covering each defect:

- the recorded transcript above, asserted to match under the real skill name and to be **missed** under a probe name — the regression for defect 1
- a real subprocess pipe read to completion through `iter_process_lines`, plus deadline behaviour — defect 2
- a `SKILL.md` containing em dashes, curly quotes and an arrow, parsed end to end; plus a scan asserting no file IO relies on the default encoding — defect 3

The encoding scan fails against the current tree with 29 offending lines.

`tests/` is excluded from packaged `.skill` bundles alongside `evals/`, so this does not add weight to what ships.

## Verification

- 12/12 tests pass on Windows 11, Python 3.12.10
- `python -m compileall` clean across `scripts/` and `eval-viewer/`
- `quick_validate.py skills/skill-creator` passes
- End to end: a 2-query eval against an installed skill returns `1.00` for the should-trigger case and `0.00` for the should-not-trigger case. Before these changes the same eval returned `0.00` for both.

I have not been able to test on macOS or Linux. Defects 2 and 3 are Windows-only by nature; defect 1 and the test suite are platform-independent.

### DIFF
--- a/skills/skill-creator/SKILL.md
+++ b/skills/skill-creator/SKILL.md
@@ -393,6 +393,31 @@ While it runs, periodically tail the output to give the user updates on which it
 
 This handles the full optimization loop automatically. It splits the eval set into 60% train and 40% held-out test, evaluates the current description (running each query 3 times to get a reliable trigger rate), then calls Claude to propose improvements based on what failed. It re-evaluates each new description on both train and test, iterating up to 5 times. When it's done, it opens an HTML report in the browser showing the results per iteration and returns JSON with `best_description` — selected by test score rather than train score to avoid overfitting.
 
+### Evaluating a skill that is already installed
+
+By default the eval writes a uniquely-named copy of the skill into
+`.claude/commands/` and checks whether Claude invoked *that* name. If the skill
+under test is already on the load path, Claude invokes the real skill instead,
+the unique name never appears, and every query is scored as a miss - the run
+completes normally and reports a 0.0 trigger rate across the board.
+
+Pass `--use-installed` to match the installed skill by its real name:
+
+```bash
+python -m scripts.run_eval \
+  --eval-set <path-to-trigger-eval.json> \
+  --skill-path <path-to-skill> \
+  --use-installed
+```
+
+This measures the description the skill currently ships with, so it is the right
+mode for auditing skills already in use. It cannot be combined with
+`--description`, because a candidate description is injected through the probe
+copy that `--use-installed` skips; passing both is rejected rather than silently
+ignoring the override. For the same reason `run_loop.py`, which tests candidate
+descriptions, always uses the probe mechanism - uninstall or temporarily move the
+skill out of the load path before optimizing it.
+
 ### How skill triggering works
 
 Understanding the triggering mechanism helps design better eval queries. Skills appear in Claude's `available_skills` list with their name + description, and Claude decides whether to consult a skill based on that description. The important thing to know is that Claude only consults skills for tasks it can't easily handle on its own — simple, one-step queries like "read this PDF" may not trigger a skill even if the description matches perfectly, because Claude can handle them directly with basic tools. Complex, multi-step, or specialized queries reliably trigger skills when the description matches.
--- a/skills/skill-creator/eval-viewer/generate_review.py
+++ b/skills/skill-creator/eval-viewer/generate_review.py
@@ -91,7 +91,7 @@ def build_run(root: Path, run_dir: Path) -> dict | None:
     for candidate in [run_dir / "eval_metadata.json", run_dir.parent / "eval_metadata.json"]:
         if candidate.exists():
             try:
-                metadata = json.loads(candidate.read_text())
+                metadata = json.loads(candidate.read_text(encoding="utf-8"))
                 prompt = metadata.get("prompt", "")
                 eval_id = metadata.get("eval_id")
             except (json.JSONDecodeError, OSError):
@@ -104,7 +104,7 @@ def build_run(root: Path, run_dir: Path) -> dict | None:
         for candidate in [run_dir / "transcript.md", run_dir / "outputs" / "transcript.md"]:
             if candidate.exists():
                 try:
-                    text = candidate.read_text()
+                    text = candidate.read_text(encoding="utf-8")
                     match = re.search(r"## Eval Prompt\n\n([\s\S]*?)(?=\n##|$)", text)
                     if match:
                         prompt = match.group(1).strip()
@@ -131,7 +131,7 @@ def build_run(root: Path, run_dir: Path) -> dict | None:
     for candidate in [run_dir / "grading.json", run_dir.parent / "grading.json"]:
         if candidate.exists():
             try:
-                grading = json.loads(candidate.read_text())
+                grading = json.loads(candidate.read_text(encoding="utf-8"))
             except (json.JSONDecodeError, OSError):
                 pass
             if grading:
@@ -222,7 +222,7 @@ def load_previous_iteration(workspace: Path) -> dict[str, dict]:
     feedback_path = workspace / "feedback.json"
     if feedback_path.exists():
         try:
-            data = json.loads(feedback_path.read_text())
+            data = json.loads(feedback_path.read_text(encoding="utf-8"))
             feedback_map = {
                 r["run_id"]: r["feedback"]
                 for r in data.get("reviews", [])
@@ -255,7 +255,7 @@ def generate_html(
 ) -> str:
     """Generate the complete standalone HTML page with embedded data."""
     template_path = Path(__file__).parent / "viewer.html"
-    template = template_path.read_text()
+    template = template_path.read_text(encoding="utf-8")
 
     # Build previous_feedback and previous_outputs maps for the template
     previous_feedback: dict[str, str] = {}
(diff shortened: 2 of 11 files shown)


98zc5g5jyw-arch:
## Babysit review — duplicate of #1630 (same author)

#1630 (opened later, same 11 files, superset) contains everything here plus the `--scan-full-turn` feature and 280 lines of tests. These two PRs edit the same `run_eval.py` hunks and cannot both merge. Suggest closing this one and keeping #1630.

98zc5g5jyw-arch (changes requested):
## Verdict: Request changes

Solid, well-tested PR — all three defects are real, and the fixes for #2 (`select` portability) and #3 (encoding) are correct and complete. I verified the encoding fix by running the PR's own scan against `main` (29 offenders) and against this branch (0 offenders); all 12 tests pass on Python 3.12; `py_compile` is clean across `scripts/` and `eval-viewer/`. The blocking issue: the PR's headline symptom — "the run completes and reports a 0.0 trigger rate" — still reproduces in the default probe mode on current Claude Code, because `.claude/commands/` mounts never surface in `available_skills`.

## Critical

**1. Default probe mode still measures nothing on Claude Code ≥ 2.1.x — must coordinate with #1609 (scripts/run_eval.py:167-247, SKILL.md:396-422).**
I reproduced this empirically with Claude Code 2.1.220: a probe mounted at `.claude/commands/zap-probe-*.md` appears in the `system/init` event's `slash_commands` array but **not** in its `skills` (available_skills) array. The model therefore never invokes the Skill tool with the probe name, and every query is scored as a miss — the exact "0.0 trigger rate" failure this PR set out to fix. Open PR #1609 makes the same claim (measured on 2.1.234) and fixes it by mounting the probe under `.claude/skills/<name>/SKILL.md` instead, which I confirmed does appear in `available_skills`.

None of this PR's end-to-end verification exercised the default path: the 2-query e2e used `--use-installed`, which skips the probe write entirely. `run_loop.py` (the optimization workflow SKILL.md documents as the primary path) still uses the probe and will report 0.0 for every candidate description on current Claude Code.

Please either: (a) adopt the `.claude/skills/` mount as in #1609 and coordinate so only one of these lands; or (b) verify which Claude Code versions surface `.claude/commands/` files in `available_skills` and document that constraint prominently in SKILL.md and in `run_eval`'s `--help`. In either case, verify the default path end-to-end on your target version — the PR's own verification only covers `--use-installed`.

## Warnings

**2. Windows console encoding for stdout is still unhandled.** `print(json.dumps(output, indent=2))` at scripts/run_eval.py:381 (and `print(json_output)` in run_loop.py's `main()`) will raise `UnicodeEncodeError` on a cp1252 console when a description contains non-ASCII; `print(html_output)` at scripts/generate_report.py:322 has the same problem, and `json.load(sys.stdin)` at generate_report.py:312 inherits the locale encoding for piped JSON. Suggest `sys.stdout.reconfigure(encoding="utf-8")` at the top of each `main()` (or a `PYTHONUTF8` note in SKILL.md).

**3. `--use-installed` + substring matching is collision-prone (scripts/run_eval.py:49-130).** With the probe, `clean_name` carries a unique suffix so accidental substring hits are unlikely. With the real (short) name, `"git" in tool_input.get("skill", "")` can match a *different* installed skill whose name contains it, and a `Read` of any file whose path contains the name counts as a trigger. Suggest exact-match on the parsed field for Skill calls (`tool_input.get("skill") == clean_name`), and documenting the Read-path heuristic.

**4. Triple-duplicated encoding work with open PRs #1591 and #1596.** All three PRs make byte-identical `encoding="utf-8"` edits to the same lines of the same six scripts. This PR's coverage is a superset (it also fixes eval-viewer/generate_review.py, aggregate_benchmark.py, package_skill.py), so please coordinate to close/supersede #1591/#1596 and avoid three PRs racing on the same hunks.

## Suggestions

- `iter_process_lines` (scripts/run_eval.py:116-152): if a caller ever passes a non-`BufferedReader`, `stream.read1` raises `AttributeError`, which the per-query catch in `run_eval` converts into a silent miss (a 0.0 contribution). Consider letting non-EOF errors propagate as a hard failure instead of a quiet `False`, or asserting the stream type.
- The PR body describes defect 3's failure mode as "reports well-formed JSON with wrong numbers", but an unhandled `UnicodeDecodeError` actually hard-crashes the run before any JSON is produced. The fix is correct either way; the framing just understates it.
- The encoding regression scan skips any line containing `encoding=`. Consider also flagging `json.load(sys.stdin)` and bare `print()` of JSON/HTML so the Windows console issue in Warning 2 is caught by the same test.

## Looks good

- `resolve_probe_name` + `--use-installed` (run_eval.py:156-187, 331-348): correct; the `--description` conflict is properly rejected via `parser.error`; probe cleanup is correctly guarded by `not use_installed`.
- The `select` → reader-thread + queue replacement preserves timeout semantics and EOF handling; both pipe tests exercise a real subprocess.
- 29 → 0 encoding offenders verified; `package_skill.py` excluding `tests/` from bundles is the right call.
- SKILL.md documentation of `--use-installed` and the `--description` constraint is clear and accurate.

PR #1630, opened 2026-08-22
skill-creator: add --scan-full-turn, so evals measure whether a skill fires at all rather than whether it fires first

nic21vdw:
Follow-up to #1629, stacked on that branch — the diff here is one commit; please read #1629 first.

A fourth way the eval harness reports a plausible number that is wrong.

## The defect

`TriggerDetector` abandons detection at the first tool call that is not `Skill` or `Read`:

```python
    pending_tool_name = tool_name
    accumulated_json = ""
else:
    return False
```

So the eval measures whether a skill fires **first**, not whether it fires at all. Any turn where Claude inspects something — a `git status`, a `Glob`, reading a file — before reaching for the skill is scored as a miss. The run completes and prints a clean trigger rate.

## How much it understates

Seven installed skills, the same 20-query eval sets, 3 runs per query, `--use-installed`, changing nothing but this flag:

| skill | early exit | `--scan-full-turn` | should-not-trigger held |
| --- | --- | --- | --- |
| blender-game-dev | 4/10 | **10/10** | 10/10 -> 10/10 |
| ship-pr | 2/10 | **7/10** | 10/10 -> 10/10 |
| doc-notes | 2/10 | **8/10** | 10/10 -> 10/10 |
| git-rescue | 8/10 | **10/10** | 10/10 -> 10/10 |
| deepseek | 8/10 | 9/10 | 10/10 -> 10/10 |
| collateral-style | 8/10 | **10/10** | 10/10 -> 9/10 |
| stream-stack | 10/10 | 10/10 | 10/10 -> 8/10 |

The queries that flipped are the ones you would predict - the ones that require looking at something before answering:

```
blender-game-dev  miss -> hit  "can you run blender without opening the gui and tell me if the scene still loads"
blender-game-dev  miss -> hit  "bake the lighting for this scene and export the textures at 2k"
blender-game-dev  miss -> hit  "package the game into something I can hand someone to run on windows"
git-rescue        miss -> hit  "did a hard reset by mistake and an hour of work in build_volume.py is gone. is it actually gone"
git-rescue        miss -> hit  "I wiped some untracked config files by cleaning the working tree. any chance of recovery"
```

`blender-game-dev` looked like a skill with a serious triggering problem. It does not have one - it scores 10/10. I had already rewritten its description on the strength of the 4/10 reading before discovering the measurement was at fault, which is the concrete cost of this defect: it sends you off to fix a description that was never broken.

### It is not free on the precision side

Five of the seven held their should-not-trigger score exactly. Two did not:

```
stream-stack       1.00  "generate thumbnails for the last 5 stream vods"
stream-stack       1.00  "write a bot in python that posts to discord when I go live"
collateral-style   0.67  "draft a cold email to a structural engineering firm about CoLateral"
```

I would call these arguable rather than clearly wrong - a stream-rig skill getting consulted about VOD thumbnails is defensible, and my eval set drew those lines. But scanning the whole turn does mean a skill consulted late, for a tangential reason, now counts as a trigger, and on this evidence that costs roughly 0-2 near-miss holds per skill. That is the tradeoff, and it is a second reason for the flag to be opt-in rather than the default.

## A caveat on my own evidence

I first tried to demonstrate this by capturing transcripts and looking for a late `Skill` call. Six captures across two skills all showed `Skill` at position 0, which pointed the other way. That sample was biased — I had chosen queries that were already passing, so of course they fired first. The flag comparison above is the reliable measurement, and it is what changed my mind. Flagging it because the constructed fixture in the tests reflects it.

## The fix

`--scan-full-turn` keeps feeding the detector until the turn ends rather than bailing at the first foreign tool. Specifically, in that mode a non-`Skill`/`Read` tool clears pending state and continues, a non-matching `content_block_stop` continues, and `message_stop` no longer terminates — detection ends at the `result` event or at the deadline.

**Off by default, deliberately.** The early exit is what lets `run_single_query` kill the subprocess as soon as the first tool block resolves; scanning the full turn runs every query to completion, which is markedly slower and costs more tokens. Users should raise `--timeout` alongside it — the 30s default cuts most turns short, and a truncated turn reads as a miss.

## Tests

Six new tests, 18 total, standard library only:

```bash
cd skills/skill-creator && python -m unittest discover -s tests
```

- default mode misses a skill invoked after another tool, `--scan-full-turn` finds it
- a turn with no `Skill` call is still a miss in full-turn mode (no false positives from just scanning longer)
- full-turn mode does not match a different skill name
- a first-position call is detected identically in both modes
- default-mode behaviour on the recorded transcript from #1629 is unchanged

The `BASH_THEN_SKILL` fixture is assembled from the event shapes in #1629's recorded transcript, reordered so the `Skill` call lands after an inspection step. It is labelled in the test file as constructed rather than recorded, for the reason in the caveat above.

## Verification

18/18 pass on Windows 11, Python 3.12.10. Default-mode results are unchanged — the 12 tests from #1629 pass untouched. The table above is the end-to-end check: 420 full-turn evals across seven installed skills, clean stderr throughout. Note the commit message quotes only the first two skills I measured, before I had run the other five; the table here is the complete picture, including the precision cost the commit message does not mention.

### DIFF
--- a/skills/skill-creator/SKILL.md
+++ b/skills/skill-creator/SKILL.md
@@ -393,6 +393,31 @@ While it runs, periodically tail the output to give the user updates on which it
 
 This handles the full optimization loop automatically. It splits the eval set into 60% train and 40% held-out test, evaluates the current description (running each query 3 times to get a reliable trigger rate), then calls Claude to propose improvements based on what failed. It re-evaluates each new description on both train and test, iterating up to 5 times. When it's done, it opens an HTML report in the browser showing the results per iteration and returns JSON with `best_description` — selected by test score rather than train score to avoid overfitting.
 
+### Evaluating a skill that is already installed
+
+By default the eval writes a uniquely-named copy of the skill into
+`.claude/commands/` and checks whether Claude invoked *that* name. If the skill
+under test is already on the load path, Claude invokes the real skill instead,
+the unique name never appears, and every query is scored as a miss - the run
+completes normally and reports a 0.0 trigger rate across the board.
+
+Pass `--use-installed` to match the installed skill by its real name:
+
+```bash
+python -m scripts.run_eval \
+  --eval-set <path-to-trigger-eval.json> \
+  --skill-path <path-to-skill> \
+  --use-installed
+```
+
+This measures the description the skill currently ships with, so it is the right
+mode for auditing skills already in use. It cannot be combined with
+`--description`, because a candidate description is injected through the probe
+copy that `--use-installed` skips; passing both is rejected rather than silently
+ignoring the override. For the same reason `run_loop.py`, which tests candidate
+descriptions, always uses the probe mechanism - uninstall or temporarily move the
+skill out of the load path before optimizing it.
+
 ### How skill triggering works
 
 Understanding the triggering mechanism helps design better eval queries. Skills appear in Claude's `available_skills` list with their name + description, and Claude decides whether to consult a skill based on that description. The important thing to know is that Claude only consults skills for tasks it can't easily handle on its own — simple, one-step queries like "read this PDF" may not trigger a skill even if the description matches perfectly, because Claude can handle them directly with basic tools. Complex, multi-step, or specialized queries reliably trigger skills when the description matches.
--- a/skills/skill-creator/eval-viewer/generate_review.py
+++ b/skills/skill-creator/eval-viewer/generate_review.py
@@ -91,7 +91,7 @@ def build_run(root: Path, run_dir: Path) -> dict | None:
     for candidate in [run_dir / "eval_metadata.json", run_dir.parent / "eval_metadata.json"]:
         if candidate.exists():
             try:
-                metadata = json.loads(candidate.read_text())
+                metadata = json.loads(candidate.read_text(encoding="utf-8"))
                 prompt = metadata.get("prompt", "")
                 eval_id = metadata.get("eval_id")
             except (json.JSONDecodeError, OSError):
@@ -104,7 +104,7 @@ def build_run(root: Path, run_dir: Path) -> dict | None:
         for candidate in [run_dir / "transcript.md", run_dir / "outputs" / "transcript.md"]:
             if candidate.exists():
                 try:
-                    text = candidate.read_text()
+                    text = candidate.read_text(encoding="utf-8")
                     match = re.search(r"## Eval Prompt\n\n([\s\S]*?)(?=\n##|$)", text)
                     if match:
                         prompt = match.group(1).strip()
@@ -131,7 +131,7 @@ def build_run(root: Path, run_dir: Path) -> dict | None:
     for candidate in [run_dir / "grading.json", run_dir.parent / "grading.json"]:
         if candidate.exists():
             try:
-                grading = json.loads(candidate.read_text())
+                grading = json.loads(candidate.read_text(encoding="utf-8"))
             except (json.JSONDecodeError, OSError):
                 pass
             if grading:
@@ -222,7 +222,7 @@ def load_previous_iteration(workspace: Path) -> dict[str, dict]:
     feedback_path = workspace / "feedback.json"
     if feedback_path.exists():
         try:
-            data = json.loads(feedback_path.read_text())
+            data = json.loads(feedback_path.read_text(encoding="utf-8"))
             feedback_map = {
                 r["run_id"]: r["feedback"]
                 for r in data.get("reviews", [])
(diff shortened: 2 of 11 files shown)


98zc5g5jyw-arch:
## Babysit review — merge this one, close #1629

Same 11 files as #1629 but a superset (adds `--scan-full-turn` + `TriggerDetector` + ~280 lines of regression tests using real `claude -p` transcripts). The two PRs touch the same `run_eval.py` hunks and will conflict — please keep this one and close #1629. Minor: the `--use-installed`/`--description` exclusivity is only enforced at the CLI layer, not in SKILL.md guidance.

98zc5g5jyw-arch (changes requested):
## Verdict: Request changes

The `--scan-full-turn` state machine is implemented correctly — I traced every branch, and all 18 tests (6 new + 12 inherited) pass on Python 3.12, with default-mode behavior provably unchanged. The blocking issues are (1) inherited from the stack: this branch sits on #1629, whose default probe path doesn't work on current Claude Code, so the flag cannot help default-mode users until that is resolved; and (2) the flag is undocumented in SKILL.md.

## Critical

**1. Inherits #1629's default-mode blindness — all evidence in this PR was produced with `--use-installed`.** I verified with Claude Code 2.1.220 that a probe mounted at `.claude/commands/*.md` appears only in `slash_commands`, never in `skills` (available_skills), so the default path scores every query a miss. Every number in this PR's table (420 evals across 7 installed skills) came from `--use-installed` runs. Until the mount mechanism is fixed (see #1609, which mounts under `.claude/skills/`), `--scan-full-turn` is a no-op for default-mode users. Please resolve the mounting question together with #1629/#1609 before landing this.

**2. Branch stacking: merging this PR merges #1629's commits too.** `fix/detect-skill-invoked-after-other-tools` contains all three commits of #1629's branch plus one (`7be6888`). If #1629 is rejected or reworked, this branch must be rebased; if both are approved, merge #1629 first so this PR's diff shrinks to its single commit. Please state the stacking explicitly in the PR description for the merge queue.

## Warnings

**3. `--scan-full-turn` is not documented in SKILL.md** (verified: at this branch's head, the only new flag documentation is #1629's `--use-installed` section). The flag's semantics ("fires at all" vs "fires first"), the precision tradeoff (this PR's own held-out data shows 1.00 → 0.80 for stream-stack and 1.00 → 0.67 for collateral-style), and the `--timeout` requirement all belong in SKILL.md next to the `--use-installed` section.

**4. Help text overpromises: "each query runs to completion" (scripts/run_eval.py:343-350) is false with the default 30s timeout.** As the PR body itself notes, the 30s default "cuts most turns short, and a truncated turn reads as a miss" — so the flag with default settings produces systematically depressed rates. Consider auto-raising the timeout floor when `--scan-full-turn` is set (e.g. ≥120s) or emitting a stderr warning, and say so in the help text.

**5. Full-turn scanning amplifies substring-match false positives (scripts/run_eval.py:60-95).** Over a whole turn, more non-Skill tool calls accumulate: a `Read` whose `file_path` contains the skill name, or another skill's `args` mentioning the tested name, now count as triggers — this is the mechanism behind the precision cost in the PR's own table. Suggest exact-match on the parsed `skill` field for Skill calls, and consider dropping the Read-path heuristic in full-turn mode (or documenting the caveat next to the flag).

**6. The `BASH_THEN_SKILL` fixture is constructed, not recorded (tests/test_run_eval.py:917-925)** — acknowledged in the PR body, but the core premise ("Claude inspects first, then reaches for the skill") rests on that fixture plus the author's own flag-comparison table; the author's six recorded captures all showed Skill at position 0. A single real recorded transcript with a late Skill call would make the defect airtight for maintainers. Also note the fixture omits `tool_result`/`user` events that real multi-turn streams interleave; the detector ignores them correctly, but that path is untested.

## Suggestions

- In full-turn mode a positive match still returns at the first matching delta (run_eval.py:80-82) and the subprocess is killed early — correct for hit/miss purposes, but "full turn" is only literal for misses; keep that nuance out of user-facing docs.
- Consider a test for the multi-turn interleaving shape (tool_result/user events between assistant turns) since real streams contain them.

## Looks good

- Default-mode behavior is provably unchanged: the 12 tests inherited from #1629 pass untouched, and the `content_block_start` / `content_block_stop` / `message_stop` / `result` semantics in default mode are bit-identical to #1629.
- The mode is opt-in, which is the right call given the measured precision cost.
- Test coverage is thoughtful: no-skill-turn miss, wrong-skill no-match, first-position equivalence, and the inherited-transcript regression.
- Flag plumbing (`run_single_query` → `run_eval` → `main`) is correct, and `--scan-full-turn` is orthogonal to `--use-installed` and `--description` — no conflicts.

PR #1632, opened 2026-08-22
docs: fix README reference to nonexistent template-skill folder

juliaamendes-go:
## Summary
- The README's "Creating a Basic Skill" section told readers to use a **template-skill** folder as a starting point, but the repository's actual folder is named `template` (see [`./template`](./template)). This PR fixes the reference and links it, matching the naming already used in the "Skill Sets" section above.

## Test plan
- [x] Verified `template/` exists and `template-skill/` does not
- [x] Confirmed the link renders correctly relative to the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ You can use Anthropic's pre-built skills, and upload custom skills, via the Clau
 
 # Creating a Basic Skill
 
-Skills are simple to create - just a folder with a `SKILL.md` file containing YAML frontmatter and instructions. You can use the **template-skill** in this repository as a starting point:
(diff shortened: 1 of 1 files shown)


98zc5g5jyw-arch:
## Babysit review — LGTM ✅

Verified: the repo root does contain `template/` (via contents API), while `template-skill` does not exist — so the README reference was indeed broken and this +1/-1 fixes it. No risk.

98zc5g5jyw-arch (approved):
## Hermes Review

**Verdict: Approve** — verified the fix against the repo.

### Verified
- `./template` exists in the repository root and contains a `SKILL.md` — the link target is real, and the previous `template-skill` folder does not exist. Correct fix.

### Suggestion (non-blocking)
- The linked text could keep the original phrasing for searchability: `[template-skill](./template)` reads naturally and preserves the old anchor text for users who search the README for "template-skill".

Issue #1624, opened 2026-08-21
claude-api: SKIP grep is unanchored — `cohere` matches "coherent", disabling the skill in any repo that uses the word


MarbTrek:
## Summary

The `claude-api` skill's SKIP clause uses an **unanchored** grep. `cohere` is a substring of the ordinary English word **"coherent"**, so the skill declines to load in any repository whose prose happens to contain that word — including repositories with no LLM integration at all.

Because SKIP "overrides all triggers", the effect is that the reference whose own description says *"never answer from memory"* is silently unavailable exactly when it is needed.

## The clause

`skills/claude-api/SKILL.md`, frontmatter `description`, line 6:

> SKIP only when another provider is being worked on (overrides all triggers): OpenAI/GPT/Gemini/Llama/Mistral/Cohere/Ollama named in the query; OR `grep -rE 'openai|langchain_openai|google.generativeai|genai|mistralai|cohere|ollama'` over the project hits (run this grep FIRST if no provider named — don't Read the file).

## Reproduction

```bash
mkdir /tmp/coherent-demo && cd /tmp/coherent-demo
echo 'This module keeps the cache in a coherent state across restarts.' > README.md

grep -rE 'openai|langchain_openai|google.generativeai|genai|mistralai|cohere|ollama' .
# ./README.md:This module keeps the cache in a coherent state across restarts.
```

The grep hits, so per the documented contract `claude-api` must SKIP — in a directory containing one sentence of English and no code.

## Scale

Measured across six unrelated private repositories (excluding `node_modules`, `.git`, `target`, `dist`, `.turbo`). "Word-anchored" is the same alternation with word boundaries applied:

| Repo | Files matching, as written | With `grep -rEw` | Disabled by "coherent" alone |
|---|---|---|---|
| A | 9 | **0** | yes |
| B | 2 | **0** | yes |
| C | 9 | **0** | yes |
| D | 1 | **0** | yes |
| E | 433 | 348 | no — genuinely uses Ollama |
| F | 40 | 16 | no — but every hit is prose *about* another repo |

**Four of six were disabled by one English word.**

Per token, `cohere` is the only alternative that matches all six repos, and in those four it is the only one that matches at all:

| Token | A | B | C | D | E | F |
|---|---|---|---|---|---|---|
| `openai`, `langchain_openai`, `google.generativeai`, `genai`, `mistralai` | 0 | 0 | 0 | 0 | 0 | 3 |
[...]

PR #1634, opened 2026-08-22
fix(claude-api): word-anchor SKIP grep pattern to prevent false positives on 'coherent' (#1624)

245678000000:
Resolves #1624

### Summary
The `claude-api` skill's SKIP clause used an unanchored `grep -rE`, causing `cohere` to match common English prose words such as `coherent`, `coherence`, or `coherently`. This caused false-positive SKIP triggers in non-LLM repositories or projects mentioning cache coherence.

Additionally, the prose description named Gemini (`OpenAI/GPT/Gemini/Llama/Mistral/Cohere/Ollama`), but `gemini` was missing from the regex alternation pattern.

### Changes
- Updated frontmatter grep command from `grep -rE` to `grep -rEw` to enforce word boundaries.
- Added `gemini` into the alternation pattern to align with the prose specification:
  `'openai|langchain_openai|google.generativeai|genai|gemini|mistralai|cohere|ollama'`

### Verification
Tested against various scenarios:
- Text containing `coherent` / `coherence` (e.g. "This module keeps the cache in a coherent state across restarts."):
  - Old `grep -rE`: Matched (false positive SKIP)
  - New `grep -rEw`: **No match (Fixed)**
- Actual provider code:
  - `import openai` / `openai.OpenAI()`: **Matched**
  - `import cohere` / `cohere.Client()`: **Matched**
  - `import gemini` / `from google.generativeai import ...`: **Matched**

### DIFF
--- a/skills/claude-api/SKILL.md
+++ b/skills/claude-api/SKILL.md
@@ -3,7 +3,7 @@ name: claude-api
 description: |-
   Reference for the Claude API / Anthropic SDK — model ids, pricing, params, streaming, tool use, MCP, agents, caching, token counting, model migration.
   TRIGGER — read BEFORE opening the target file; don't skip because it "looks like a one-liner" — whenever: the prompt names Claude/Anthropic in any form (Claude, Anthropic, Fable, Opus, Sonnet, Haiku, `anthropic`, `@anthropic-ai`, `claude-*`, `us.anthropic.*`, `[1m]`); the user asks about an LLM (pricing/model choice/limits/caching) — never answer from memory; OR the task is LLM-shaped with provider unstated (agent/MCP/tool-definition/multi-agent/RAG/LLM-judge/computer-use; generate/summarize/extract/classify/rewrite/converse over NL; debugging refusals/cutoffs/streaming/tool-calls/tokens).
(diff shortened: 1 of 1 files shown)


98zc5g5jyw-arch:
## Babysit review — duplicate of #1636

FYI: #1636 (opened later) makes a **semantically identical** one-line change to the same SKIP grep line (both add `-rEw` word anchoring + `gemini` token for #1624). Only the alternation order differs. Suggest coordinating with that author so only one of the two lands. The change itself is correct and I've verified it resolves both defects in #1624.

98zc5g5jyw-arch (commented):
## Hermes Review

**Verdict: Comment** — change is correct; **duplicate of PR #1636** (identical edit to the same line).

### Verified
- `-w` word anchoring fixes the #1624 false positive on `coherent` (no longer matches `cohere`), while `langchain_openai` / `google.generativeai` / `mistralai` / `cohere` / `ollama` still match correctly as whole words.
- `gemini` was genuinely missing from the grep pattern — good addition.

### Suggestion
- PR #1636 makes the **exact same change**. Coordinate with the author: one should be closed so the other merges cleanly. Suggest rebasing whichever survives.

PR #1635, opened 2026-08-23
fix(claude-api): trim description to 986 chars to fit 1024-char frontmatter cap

volcanicll:
Fixes #1622

## Summary

The `description` frontmatter in `skills/claude-api/SKILL.md` was 1,077 characters, exceeding the 1,024-character cap enforced by some skill loaders (e.g. pi.dev). This PR trims it to **986 characters** while preserving all trigger and skip semantics.

## Changes

All edits are pure compression — no trigger, skip, or provider-detection behavior is altered:

| Line | Before | After | Saved |
|------|--------|-------|-------|
| TRIGGER | "names Claude/Anthropic in any form (" | "names Claude/Anthropic (" | 10 |
| TRIGGER | "the user asks" / "the task is" | "user asks" / "task is" | 8 |
| TRIGGER | "(agent/MCP/tool-definition/multi-agent/RAG/LLM-judge/computer-use; generate/summarize/extract/classify/rewrite/converse over NL; debugging refusals/cutoffs/streaming/tool-calls/tokens)" | "(agents, MCP, tool use, RAG, LLM-judge, computer-use; generating/summarizing/classifying/rewriting NL; debugging refusals, cutoffs, tool calls, token issues)" | 9 |
| SKIP | "SKIP only when another provider is being worked on" | "SKIP when another provider is in play" | 14 |
| SKIP | "if no provider named — don't Read the file)" | "if no provider named)" | 24 |

## Validation

- [x] Description length: **986 / 1024** (was 1077)
- [x] YAML frontmatter parses correctly (`yaml.safe_load`)
- [x] All model-ID aliases preserved (`claude-*`, `us.anthropic.*`, `[1m]`, etc.)
- [x] All skip-grep patterns preserved verbatim
- [x] All LLM-shaped task categories preserved

PR #1636, opened 2026-08-23
fix(claude-api): word-anchor skip grep and add missing gemini token

volcanicll:
Fixes #1624

## Summary

Two defects in the `claude-api` skill's SKIP-clause grep, both reported in #1624:

### 1. False positive: unanchored `cohere` matches the English word "coherent"

`grep -rE 'cohere'` fires on any prose containing *coherent/coherence/coherently*. Because SKIP overrides all triggers, the skill silently disables itself in repos that merely contain ordinary English. Fix: `-rEw` (word-anchored), portable across GNU and BSD grep.

### 2. False negative: Gemini named in prose but absent from the pattern

The SKIP clause names Gemini as a provider but the grep had no `gemini` token — repos genuinely integrating Gemini via SDK were not detected by the SKIP check. Fix: added `gemini` to the alternation.

## Verification

```bash
# coherent no longer matches:
echo 'This module keeps the cache in a coherent state.' > a.md
grep -rlEw '...|cohere|...' .   # → (no match)  ✓

# genuine imports still match:
echo 'import openai' > b.py
echo 'from google import genai' > c.py
echo 'use gemini api' > d.txt
grep -rlEw 'openai|...|gemini|ollama' .   # → b.py c.py d.txt  ✓
```

## Note

This is a minimal fix scoped exactly to what #1624 identifies. The issue's broader design observation (frontmatter repo-wide grep vs the body's per-file scan in `## Before You Start`) is separable and left untouched.

### DIFF
--- a/skills/claude-api/SKILL.md
+++ b/skills/claude-api/SKILL.md
@@ -3,7 +3,7 @@ name: claude-api
 description: |-
   Reference for the Claude API / Anthropic SDK — model ids, pricing, params, streaming, tool use, MCP, agents, caching, token counting, model migration.
   TRIGGER — read BEFORE opening the target file; don't skip because it "looks like a one-liner" — whenever: the prompt names Claude/Anthropic in any form (Claude, Anthropic, Fable, Opus, Sonnet, Haiku, `anthropic`, `@anthropic-ai`, `claude-*`, `us.anthropic.*`, `[1m]`); the user asks about an LLM (pricing/model choice/limits/caching) — never answer from memory; OR the task is LLM-shaped with provider unstated (agent/MCP/tool-definition/multi-agent/RAG/LLM-judge/computer-use; generate/summarize/extract/classify/rewrite/converse over NL; debugging refusals/cutoffs/streaming/tool-calls/tokens).
-  SKIP only when another provider is being worked on (overrides all triggers): OpenAI/GPT/Gemini/Llama/Mistral/Cohere/Ollama named in the query; OR `grep -rE 'openai|langchain_openai|google.generativeai|genai|mistralai|cohere|ollama'` over the project hits (run this grep FIRST if no provider named — don't Read the file).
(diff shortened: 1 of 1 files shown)


98zc5g5jyw-arch:
## Babysit review — duplicate of #1634

This PR and #1634 make **semantically identical** changes to the same line in `skills/claude-api/SKILL.md` (both fix #1624): add word-anchoring `-rEw` to the SKIP grep and add the missing `gemini` token. The only difference is alternation order:

- #1634: `...|genai|gemini|mistralai|cohere|ollama`
- #1636: `...|genai|mistralai|cohere|gemini|ollama`

Verified via `git diff pr-1634 pr-1636 -- skills/claude-api/SKILL.md` — one-line diff, no semantic difference.

**Suggested action:** coordinate with the author of #1634 (opened ~12h earlier). Close one of the two (e.g. keep the earlier one and reference it here, or vice versa) so maintainers don't have to reconcile two competing patches for the same issue. The fix itself is correct: `-rEw` kills the `coherent` false positive and `gemini` closes the false-negative gap.

Also note: #1635 (same author) rewrites this exact SKIP line for the description-length fix, so whichever lands first will conflict with the other — worth rebasing the other on top.

98zc5g5jyw-arch (commented):
## Hermes Review

**Verdict: Comment** — change is correct; **duplicate of PR #1634** (identical edit to the same line).

### Verified
- `grep -rEw` word-anchoring fixes the #1624 false positive: `coherent` no longer matches the `cohere` alternative (word boundary required), while `langchain_openai`, `google.generativeai`, `mistralai` still match as whole words.
- Adding `gemini` to the pattern closes a real gap: the query text mentions Gemini as a SKIP trigger, but the grep could not catch bare `gemini` usage in the project.

### Suggestion
- PR #1634 makes the **exact same change** to the same line. Please coordinate: keep one, close the other (and rebase whichever stays, since only one can apply cleanly).
- Consider also testing that `-w` still matches `google.generativeai` (it does — the whole literal matches), so no regression risk.

Issue #1617, opened 2026-08-20
fill_fillable_fields.py silently overwrites existing files (no output-path check) Affected file: pdf/scripts/fill_fillable_fields.py Related doc: pdf/forms.md (lines 74–76, instructs the agent to invoke the script with a caller-supplied output path)


benedictkwok:
fill_fillable_fields.py writes to output_pdf_path unconditionally, with no check for whether a file already exists at that path:

with open(output_pdf_path, "wb") as f:
    writer.write(f)

If the output path collides with an existing file — including the original input PDF, if the same path is passed for both <input pdf> and <output pdf> — that file is silently truncated and overwritten. There is no confirmation prompt, no --force flag, and no warning to the user or the calling agent.

Reproduction
1.Have a fillable PDF at form.pdf.
2. Run: python scripts/fill_fillable_fields.py form.pdf field_values.json form.pdf
3. form.pdf is overwritten in place with the filled version — the original is unrecoverable. No error, no confirmation.

This is a plausible outcome in normal use, not just a contrived edge case: forms.md documents no default output-naming convention, so an agent following the skill's instructions could reasonably choose to "fill the form in place" absent more specific direction from the user.

Classification

CWE-73 (External Control of File Name or Path) combined with a missing overwrite guard — the output path is fully caller/agent-controlled with no existence check before the destructive write. As implemented today it's a data-integrity defect triggered by ordinary usage rather than an externally-exploitable vulnerability; we're flagging the CWE for precision, not asserting a higher severity than the facts support. It would escalate meaningfully if the output path were ever derived from untrusted or externally-influenced input (e.g., document metadata, injected instructions) rather than direct user intent.

Suggested fix

Before writing:

python
if os.path.exists(output_pdf_path) and os.path.abspath(output_pdf_path) == os.path.abspath(input_pdf_path):
    print(f"ERROR: output path matches input path ({output_pdf_path}); refusing to overwrite without --force")
    sys.exit(1)
if os.path.exists(output_pdf_path) and not force:
    print(f"ERROR: {output_pdf_path} already exists; pass --force to overwrite")
    sys.exit(1)

And update forms.md's invocation instructions to tell the agent to check for an existing output file / choose a non-colliding filename by default (e.g., append _filled) rather than leaving that entirely to caller discretion.

Reported by

[...]

PR #1639, opened 2026-08-23
fix(pdf): refuse to overwrite input file in fill_fillable_fields

Xsidz:
Fixes #1617.

## Problem

`fill_fillable_fields.py` wrote the output PDF unconditionally with no check for path collisions. Passing the same path for both input and output silently truncated and overwrote the source file — no error, no warning, no recovery.

## Fix

Added an `os.path.abspath` equality check before the write. If `output_pdf_path` resolves to the same filesystem path as `input_pdf_path`, the script exits with a clear error message before touching the file.

```python
abs_in = os.path.abspath(input_pdf_path)
abs_out = os.path.abspath(output_pdf_path)
if abs_in == abs_out:
    print(f"ERROR: output path matches input path ({output_pdf_path}); refusing to overwrite source file")
    sys.exit(1)
```

## Scope

One file changed (`skills/pdf/scripts/fill_fillable_fields.py`). Does not add a `--force` flag or block unrelated existing-file overwrites — those remain the caller's responsibility and are consistent with how the rest of the pdf skill's scripts behave.

### DIFF
--- a/skills/pdf/scripts/fill_fillable_fields.py
+++ b/skills/pdf/scripts/fill_fillable_fields.py
@@ -1,4 +1,5 @@
 import json
+import os
 import sys
 
 from pypdf import PdfReader, PdfWriter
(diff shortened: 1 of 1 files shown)


98zc5g5jyw-arch (approved):
Reviewed by Hermes Agent. Refuses to overwrite input file by comparing abspath of input and output. Correct direction. Suggestion (non-blocking): consider os.path.realpath to also cover symlink/hardlink aliases of the same file.

98zc5g5jyw-arch:

**Verdict: good, approve with one suggestion.** The abs-path equality check correctly prevents clobbering the source PDF.

Suggestion: consider `os.path.realpath` (or `os.path.samefile` on POSIX) instead of `abspath` — a symlink or hardlink pointing at the input file would pass the current check yet still overwrite the same underlying file. Cheap to harden, but the fix as-is covers the common case.

Issue #1618, opened 2026-08-20
Path traversal in the resource-handler example (python_mcp_server.md) — unsanitized URI template parameter used directly in a filesystem path  Affected file: mcp-builder/reference/python_mcp_server.md (Resource Registration section, ~line 527–541)


benedictkwok:
Summary
The example resource handler builds a file path by directly interpolating the name parameter — which comes straight from the MCP client's URI template request, i.e. caller-controlled input — into a filesystem path with no sanitization or bounds checking:

python
`@mcp.resource("file://documents/{name}")
async def get_document(name: str) -> str:
    '''Expose documents as MCP resources.

    Resources are useful for static or semi-static data that doesn't
    require complex parameters. They use URI templates for flexible access.
    '''
    document_path = f"./docs/{name}"
    with open(document_path, "r") as f:
        return f.read()
`

A client requesting file://documents/../../etc/passwd (or any other ..-containing name) would cause document_path to resolve outside the intended ./docs/ directory. There's no surrounding caveat or warning in the guide calling this out — it's presented as the standard way to register a resource.

Note on scope: this is a pattern in teaching/example content, not a live service — we have not deployed and probed a running server to confirm exploitation end-to-end, and are flagging based on direct code inspection rather than a fired PoC. The practical risk is that this guide exists specifically to be copied into real MCP server implementations by developers building their own servers, so an unsanitized example here has a real chance of propagating into production code.

Classification

CWE-22 (Improper Limitation of a Pathname to a Restricted Directory / Path Traversal).

Suggested fix
python
import os

DOCS_ROOT = os.path.realpath("./docs")

`@mcp.resource("file://documents/{name}")
async def get_document(name: str) -> str:
    document_path = os.path.realpath(os.path.join(DOCS_ROOT, os.path.basename(name)))
    if not document_path.startswith(DOCS_ROOT + os.sep):
        raise ValueError("invalid document name")
    with open(document_path, "r") as f:
        return f.read()
`
Alternatively, add an explicit inline warning in the guide noting that URI template parameters must never be used raw in filesystem operations, with a pointer to the sanitization pattern above — even if the full fix isn't adopted verbatim, flagging the risk at the point of the example would reduce copy-paste propagation.

Reported by

Benedict Kwok, ZTAI Security Advisors LLC ([benedictkwok@ztai.ai](mailto:benedictkwok@ztai.ai)) — found while independently evaluating this skill against NVIDIA SkillSpector.

PR #1640, opened 2026-08-23
fix(mcp-builder): guard resource handler example against path traversal

Xsidz:
Fixes #1618.

## Problem

The `get_document` example in `python_mcp_server.md` passed the MCP client-supplied `name` parameter directly into a file path:

```python
document_path = f"./docs/{name}"
```

A client requesting `file://documents/../../etc/passwd` would escape the `./docs/` directory entirely. Because this guide is intended to be copied into real MCP server implementations, the unsanitized pattern propagates into production code.

## Fix

Replace the bare f-string interpolation with a `realpath`-based bounds check:

```python
DOCS_ROOT = os.path.realpath("./docs")
document_path = os.path.realpath(os.path.join(DOCS_ROOT, os.path.basename(name)))
if not document_path.startswith(DOCS_ROOT + os.sep):
    raise ValueError(f"invalid document name: {name!r}")
```

`os.path.basename` strips any directory component from the input; `realpath` resolves symlinks; the `startswith` guard is the belt-and-suspenders check. The combined approach follows the pattern suggested in the issue and CWE-22 guidance.

## Scope

One file changed (`skills/mcp-builder/reference/python_mcp_server.md`). No runtime code touched.

### DIFF
--- a/skills/mcp-builder/reference/python_mcp_server.md
+++ b/skills/mcp-builder/reference/python_mcp_server.md
@@ -529,6 +529,10 @@ async def interactive_tool(resource_id: str, ctx: Context) -> str:
 Expose data as resources for efficient, template-based access:
 
 ```python
+import os
+
+DOCS_ROOT = os.path.realpath("./docs")
+
 @mcp.resource("file://documents/{name}")
 async def get_document(name: str) -> str:
     '''Expose documents as MCP resources.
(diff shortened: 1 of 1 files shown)


patelashit550-cpu (approved):

98zc5g5jyw-arch (approved):
Reviewed by Hermes Agent. Path traversal guard in docs example: os.path.realpath + basename strip + startswith(DOCS_ROOT + os.sep) boundary check. Correct defense in depth. LGTM.

patelashit550-cpu (approved):
[native-executables.json](https://github.com/user-attachments/files/31377821/native-executables.json)

98zc5g5jyw-arch:

**Verdict: good fix, approve-worthy.** `os.path.basename(name)` strips traversal components, and `realpath` + `startswith(DOCS_ROOT + os.sep)` is correct defense in depth for the doc example.

Non-blocking nit: on case-insensitive filesystems (default macOS/Windows), a name like `DoCs/../../etc/passwd` — after basename it's just `passwd`, so the prefix check is still safe here. For extra robustness `os.path.commonpath` would be case-aware, but not required for this example.

Issue #1631, opened 2026-08-22
Path traversal via unsanitized skill_name in run_eval.py


truongsontung:
## Summary

The `run_eval.py` script uses `skill_name` from SKILL.md frontmatter directly in file path construction without sanitization, allowing path traversal.

## Vulnerability

**Location:** `skills/skill-creator/scripts/run_eval.py:54`

```python
project_commands_dir = Path('~/.claude/commands')
clean_name = re.sub(r'[^a-z0-9-]', '', skill_name.lower())
command_file = project_commands_dir / f{clean_name}.md
```

The `clean_name` only strips non- characters but doesn't prevent  sequences. An attacker providing a SKILL.md with  would write the command file outside the intended directory.

**Compare with** `generate_review.py` (line 461): This file has SSRF via `webbrowser.open`, but the port validation constrains it to integer.

## Impact

- Arbitrary file write outside `~/.claude/commands/`
- Potential RCE if combined with other vulnerabilities

## Recommended Fix

1. Call the existing `quick_validate.py` validation
2. Or explicitly reject  in the name
3. Use `Path.resolve()` and check it stays within the target directory

## References

- `quick_validate.py` already has regex: ^[a-z0-9-]+$
- Related: `skills/skill-creator/scripts/run_eval.py` line 54

Alsami1998:
HI

Alsami1998:
LKPKP

PR #1642, opened 2026-08-24
fix(skill-creator): prevent path traversal in run_eval

Arjunr7019:
## What changed?

* Fixed the path traversal vulnerability in `skills/skill-creator/scripts/run_eval.py`.
* Sanitized `skill_name` before using it to construct the generated command file path.
* Added a safe fallback when sanitization produces an empty skill name.
* Added a defense-in-depth check using `Path.resolve()` to ensure the final path remains inside the intended `~/.claude/commands/` directory.

## Why?

`skill_name` is read from the `SKILL.md` frontmatter and was used to construct the command file path without sufficient validation.

This could allow a malicious skill name containing path traversal sequences to cause the generated command file to be written outside the intended `~/.claude/commands/` directory.

The issue identified this as a potential arbitrary file-write vulnerability, with possible security impact if combined with other vulnerabilities.

## How?

The fix now:

1. Sanitizes `skill_name` using the existing strict naming pattern:

```python
re.sub(r"[^a-z0-9-]", "", skill_name.lower())
```

2. Falls back to `"skill"` if sanitization removes the entire name.

3. Resolves the final command-file path and verifies that it remains within the intended commands directory before writing.

This provides both input sanitization and a filesystem boundary check.

## Testing

Tested with malicious and invalid inputs, including:

```text
../../../../etc/cron.d/evil
a/b/c
```

These inputs are now sanitized and/or rejected by the path boundary check, ensuring generated files remain inside the intended commands directory.

## Security

This prevents `run_eval.py` from writing generated command files outside `~/.claude/commands/` through a malicious `skill_name`.

Fixes #1631

### DIFF
--- a/skills/skill-creator/scripts/run_eval.py
+++ b/skills/skill-creator/scripts/run_eval.py
@@ -8,6 +8,7 @@ for a set of queries. Outputs results as JSON.
 import argparse
 import json
 import os
+import re
 import select
 import subprocess
 import sys
@@ -49,9 +50,24 @@ def run_single_query(
     full assistant message, which only arrives after tool execution.
     """
     unique_id = uuid.uuid4().hex[:8]
-    clean_name = f"{skill_name}-skill-{unique_id}"
+    # Sanitize skill_name before using it in a file path: skill_name comes
+    # from SKILL.md frontmatter, which may be attacker-controlled, so it
+    # must not be allowed to contain path separators or ".." sequences.
+    # Mirror the strict pattern already enforced in quick_validate.py.
+    safe_skill_name = re.sub(r"[^a-z0-9-]", "", skill_name.lower())
+    if not safe_skill_name:
+        safe_skill_name = "skill"
+    clean_name = f"{safe_skill_name}-skill-{unique_id}"
     project_commands_dir = Path(project_root) / ".claude" / "commands"
-    command_file = project_commands_dir / f"{clean_name}.md"
+    command_file = (project_commands_dir / f"{clean_name}.md").resolve()
+
+    # Defense in depth: verify the resolved path actually stays within the
+    # intended commands directory before writing anything.
+    resolved_commands_dir = project_commands_dir.resolve()
+    if resolved_commands_dir not in command_file.parents:
+        raise ValueError(
+            f"Refusing to write command file outside {resolved_commands_dir}: {command_file}"
+        )
 
     try:
         project_commands_dir.mkdir(parents=True, exist_ok=True)


98zc5g5jyw-arch (approved):
Reviewed by Hermes Agent. Security fix is correct and defense-in-depth:

✅ **Sanitize is sound** — `re.sub(r"[^a-z0-9-]", "", skill_name.lower())` strips all path separators (`/`, `\\`), dots (`..` sequences), and whitespace, killing every traversal vector (`../../etc/passwd` → `etcpasswd`). Allow-list set matches quick_validate.py's `^[a-z0-9-]+$` strict pattern (verified at quick_validate.py:65) — the "mirror" comment is accurate.
✅ **Empty-name fallback** — all-special-char name → `skill`, so `clean_name` is never empty and never collides with the commands dir itself.
✅ **resolve() + parents check** — `.resolve()` defeats symlink aliasing before the containment test; `resolved_commands_dir in command_file.parents` is a strict ancestor check (stronger than the prefix/startswith patterns used in #1640). ValueError is raised *before* the try/mkdir, propagates to run_eval's `except Exception` → fail-closed, no silent partial write.
✅ **Cleanup consistent** — the final `command_file.unlink()` uses the same (resolved) path object, so no stale probe files.

⚠️ **Cross-PR note for maintainers (non-blocking):** #1629 (already reviewed, APPROVED) rewrites the *same line* (`clean_name = resolve_probe_name(...)`), and that helper returns the raw `skill_name` **unsanitized** when `use_installed=False`. If #1629 merges first and #1642 is dropped or conflicts, the traversal vector returns. Suggest: merge #1642 first, or fold this sanitize into `resolve_probe_name` when resolving the conflict. Also note #1629 rewrites `run_single_query` heavily (stream-parser refactor) while #1642 touches its first 20 lines — expect a small conflict either way.

No CI on this repo (checks empty), so no CI failures to fix. Nice minimal patch — one file, 17 lines.

98zc5g5jyw-arch:

**Verdict: looks good — approve-worthy.** The fix correctly prevents path traversal via attacker-controlled SKILL.md frontmatter `skill_name`:

1. Sanitization `re.sub(r"[^a-z0-9-]", "", skill_name.lower())` mirrors the strict pattern in quick_validate.py and fully removes separators and `..` sequences — correct first line of defense.
2. `.resolve()` + `parents` containment check is proper defense in depth, and the error message names the intended directory.

Minor note (non-blocking): sanitization also strips underscores, so a skill named `my_skill` becomes `myskill` in the command filename. Acceptable for safety, just worth being aware of.

Arjunr7019:
Thanks for the detailed review and for approving the PR! I appreciate the security checks and the note about the potential conflict with #1629.

PR #1645, opened 2026-08-24
Add claude-academy plugin: the academy-guide skill plus the Claude Academy MCP connector

kswan-wk:
## What this adds

A new plugin at `plugins/claude-academy`, listed in this repo's marketplace as `claude-academy`, so one install gives Claude both the `academy-guide` skill and a connection to the Claude Academy MCP server.

- `plugins/claude-academy/.claude-plugin/plugin.json` — plugin metadata (name `claude-academy`, display name "Claude Academy", version 1.0.0, Apache-2.0).
- `plugins/claude-academy/skills/academy-guide/` — a copy of `skills/academy-guide` (SKILL.md and LICENSE.txt), identical to the top-level copy.
- `plugins/claude-academy/.mcp.json` — the Claude Academy MCP server at `https://academy.claude.com/mcp` (streamable HTTP, no sign-in, read-only tools: search, get one item, list content).
- `plugins/claude-academy/README.md` — what the plugin does, how to install it, example prompts, maintainer notes.
- `.claude-plugin/marketplace.json` — a new `claude-academy` entry pointing at `./plugins/claude-academy`. The existing `academy-guide` entry is unchanged, so anyone who already installed it is unaffected.

## Why

The `academy-guide` skill (added in #1554, updated in #1605) recommends Claude Academy courses, tutorials, and use cases. Academy now also has an MCP server that lets Claude search and read that content directly. Bundling the two as one plugin gives people a single "install" for both, and gives any plugin catalog one self-contained directory to point at.

## Notes for reviewers

- **Two copies of the skill.** `skills/academy-guide` stays the canonical copy and still backs the standalone `academy-guide` marketplace entry. This repo's own marketplace could reference the top-level skill by path, but a plugin catalog pulls a plugin as one self-contained directory, so the plugin carries its own copy. Both must be updated together; the plugin README says so.
- **`.mcp.json` is optional.** If the preference is to ship a skill-only plugin and list the MCP server separately as a connector, delete `.mcp.json` and the connector sentences in the README; nothing else depends on it.
- **Version is pinned at 1.0.0.** With a pinned version, installed copies only update when it is bumped. Drop the `version` field if every commit should count as a new version instead.

## Validation

`claude plugin validate plugins/claude-academy` and `claude plugin validate .claude-plugin/marketplace.json` both pass. `diff -r skills/academy-guide plugins/claude-academy/skills/academy-guide` is empty. Installed the plugin from this branch into a clean config via a local marketplace: `claude plugin list` shows `claude-academy` 1.0.0 enabled, and `claude mcp list` shows the `claude-academy` server at `https://academy.claude.com/mcp` as connected.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -68,6 +68,11 @@
       "skills": [
         "./skills/discernment-nudge"
       ]
+    },
+    {
+      "name": "claude-academy",
+      "description": "Claude Academy plugin: the academy-guide skill plus a connector to the Claude Academy MCP server for searching and reading Academy courses, tutorials, and use cases",
+      "source": "./plugins/claude-academy"
     }
   ]
 }
--- a/plugins/claude-academy/.claude-plugin/plugin.json
+++ b/plugins/claude-academy/.claude-plugin/plugin.json
@@ -1,0 +1,19 @@
+{
+  "name": "claude-academy",
+  "displayName": "Claude Academy",
+  "version": "1.0.0",
+  "description": "Recommends Claude Academy courses, tutorials, and use cases when you ask how to use Claude, and connects to the Claude Academy server so Claude can search and read Academy content.",
+  "author": {
+    "name": "Anthropic"
+  },
+  "homepage": "https://academy.claude.com",
+  "repository": "https://github.com/anthropics/skills",
+  "license": "Apache-2.0",
+  "keywords": [
+    "academy",
+    "learning",
+    "courses",
+    "tutorials",
+    "claude"
+  ]
+}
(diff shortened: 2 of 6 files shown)


98zc5g5jyw-arch:

**Verdict: looks good, approve-worthy.** Verified concretely:

1. `plugins/claude-academy/skills/academy-guide/SKILL.md` is **byte-identical** to the canonical top-level `skills/academy-guide/SKILL.md` (147 = 147 lines) — no drift between the plugin copy and the standalone entry.
2. Marketplace registration is consistent: entry `claude-academy` → `./plugins/claude-academy`, and the README install command `claude-academy@anthropic-agent-skills` matches the repo marketplace name (`anthropic-agent-skills`).
3. `.mcp.json` uses `"type": "http"` (streamable HTTP) — correct modern transport shape; read-only, no sign-in claim matches the design.

One maintainer note (already documented in the README, just flagging): the dual-copy pattern means every future edit to academy-guide must touch both copies — worth a CI check that compares them, or a symlink if the plugin loader supports it.

Issue #1686, opened 2026-08-28
skill-creator scripts crash on Windows for any skill containing non-ASCII text (missing encoding on read_text())


mladenaleta:
## Summary

Three scripts in `skills/skill-creator/scripts/` read or write text without
specifying an encoding. On Windows, Python defaults to the ANSI code page
(cp1252 on most systems), so any SKILL.md containing non-ASCII characters
raises `UnicodeDecodeError` before validation begins.

This affects Anthropic's own shipped skills. In a local library of 60 skills,
**28 crashed the validator (47%)**, including `docx`, `pdf`, `pptx`, `xlsx`,
`mcp-builder`, and `webapp-testing`.

The bug is invisible on macOS and Linux, where the default encoding is UTF-8.

## Reproduction

Windows, Python 3.x:

```bash
mkdir crash-demo && cd crash-demo
cat > SKILL.md <<'EOF'
---
name: crash-demo
description: Provjeri dostupnost artikla i rok isporuke.
---
Body.
EOF
cd ..
python skills/skill-creator/scripts/quick_validate.py crash-demo
```

Result:

```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8d in position 1476:
character maps to <undefined>
```

Any non-ASCII character reproduces it — accented Latin, CJK, emoji, or a
typographic dash.

## Affected locations

**1. `scripts/quick_validate.py`, line 22**
```python
content = skill_md.read_text()
```
Crashes before any validation runs.

**2. `scripts/utils.py`, line 9 (`parse_skill_md`)**
```python
content = (skill_path / "SKILL.md").read_text()
```
Same defect. Reached via `improve_description.py` and `run_eval.py`, so the
description-optimization loop fails on the same skills.

**3. `scripts/package_skill.py`, line 122 (separate but related)**
```python
print(f"\U0001f4e6 Packaging skill: {skill_path}")
```
Raises `UnicodeEncodeError` writing the 📦 emoji to a cp1252 console, so
packaging fails on Windows for *every* skill regardless of content:
```
UnicodeEncodeError: 'charmap' codec can't encode character '\U0001f4e6'
```
Confirmed workaround: `PYTHONIOENCODING=utf-8`.

## Suggested fix

For the two readers:

```python
content = skill_md.read_text(encoding="utf-8-sig")
```

`utf-8-sig` is byte-identical to `utf-8` on files without a byte-order mark
and additionally tolerates one when present. A real BOM was found in the
[...]

PR #1687, opened 2026-08-28
fix(skill-creator): specify utf-8 encoding on file operations for Windows (#1686)

loulanyue:
## Summary

Fixes #1686

### Root Cause
On Windows systems where Python defaults to ANSI/cp1252 instead of UTF-8, scripts in `skills/skill-creator/scripts/` calling `read_text()` or `write_text()` without an explicit encoding raise `UnicodeDecodeError` (e.g. `'charmap' codec can't decode byte 0x8d...`) on any SKILL.md or eval JSON containing non-ASCII text (accented Latin, CJK, dashes, or emoji). Additionally, terminal emoji logging in `package_skill.py` can crash with `UnicodeEncodeError` under Windows console encodings.

### Changes
1. **`quick_validate.py` & `utils.py`**:
   - Use `encoding='utf-8-sig'` when reading `SKILL.md`, supporting standard UTF-8 as well as files with a UTF-8 BOM.
2. **`run_eval.py`, `improve_description.py`, `run_loop.py`, `generate_report.py`**:
   - Specify `encoding='utf-8'` on `read_text()` and `write_text()` operations.
3. **`package_skill.py`**:
   - Reconfigure stdout with `sys.stdout.reconfigure(encoding="utf-8", errors="replace")` if available to protect emoji output on Windows.
   - Add fallback import handling for `validate_skill`.
4. **Unit Tests**:
   - Add `skills/skill-creator/tests/test_windows_encoding.py` covering non-ASCII characters, UTF-8 BOM, and packaging.

### Verification
- `pytest skills/skill-creator/tests/test_windows_encoding.py` passed (2/2).
- `ruff check skills/skill-creator/tests/test_windows_encoding.py` passed.

### DIFF
--- a/skills/skill-creator/scripts/generate_report.py
+++ b/skills/skill-creator/scripts/generate_report.py
@@ -311,13 +311,14 @@ def main():
     if args.input == "-":
         data = json.load(sys.stdin)
     else:
-        data = json.loads(Path(args.input).read_text())
+        data = json.loads(Path(args.input).read_text(encoding="utf-8"))
 
     html_output = generate_html(data, skill_name=args.skill_name)
 
     if args.output:
-        Path(args.output).write_text(html_output)
+        Path(args.output).write_text(html_output, encoding="utf-8")
         print(f"Report written to {args.output}", file=sys.stderr)
+
     else:
         print(html_output)
 
--- a/skills/skill-creator/scripts/improve_description.py
+++ b/skills/skill-creator/scripts/improve_description.py
@@ -205,10 +205,11 @@ def main():
         print(f"Error: No SKILL.md found at {skill_path}", file=sys.stderr)
         sys.exit(1)
 
-    eval_results = json.loads(Path(args.eval_results).read_text())
+    eval_results = json.loads(Path(args.eval_results).read_text(encoding="utf-8"))
     history = []
     if args.history:
-        history = json.loads(Path(args.history).read_text())
+        history = json.loads(Path(args.history).read_text(encoding="utf-8"))
+
 
     name, _, content = parse_skill_md(skill_path)
     current_description = eval_results["description"]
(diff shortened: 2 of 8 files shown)


Hahaknight (approved):
Verified on real Windows 11 (Python 3.13). Environment note for transparency: this machine has the system-wide UTF-8 ACP (cp65001) enabled, so I reproduced the cp936 failure mode via controlled encoding forcing rather than a native crash — using **real production skills as the corpus** (a shipped Chinese-language skill set: 3 pure-CJK SKILL.md files + 7 with CJK sections).

Results:

- **Tests**: `tests/test_windows_encoding.py` → **2/2 pass** (validate + full package with non-ASCII).
- **Corpus**: fixed `quick_validate.py` over 10 real production skills → **10/10 valid**, 0 failures.
- **BOM edge** (the reason `utf-8-sig` matters over plain `utf-8`): prepending a real UTF-8 BOM (`EF BB BF`) to a shipped Chinese SKILL.md — plain `utf-8` leaves `﻿` in place so `content.startswith('---')` returns **False** and a *valid* skill is silently rejected; `utf-8-sig` passes. `utf-8-sig` is byte-identical to `utf-8` on non-BOM files, so there's no downside.
- **Original crash**: `read_text(encoding='cp936')` on the same non-BOM file raises `UnicodeDecodeError` — confirming the report's mechanism for typical Chinese/CJK Windows users (default GBK code page).
- **package_skill stdout**: under `PYTHONIOENCODING=cp936`, printing `📦` raises `UnicodeEncodeError: 'gbk' codec can't encode`; the `sys.stdout.reconfigure(...)` approach in this PR fixes it.

This PR covers all three locations named in #1686 (`quick_validate.py`, `utils.py` `parse_skill_md`, `package_skill.py` stdout) and is the only one of the two open fixes that handles the BOM case the reporter documented finding in the wild.

Issue #1692, opened 2026-08-30
[skill-creator] Trigger-eval harness silently broken on Windows: select() on pipes, probe files leak into global ~/.claude/commands, parallel twins cap recall at ~1/N


almezaniy:
## Summary

The skill-creator description-optimization harness (`skills/skill-creator/scripts/run_eval.py` + `run_loop.py`) is structurally broken on Windows. Worse than crashing, it **fails silently**: every probe dies or mismeasures, the loop reports `recall=0%` (or a hard ~1/N cap), and then happily "improves" the description against pure noise for up to 5 iterations. Each probe is a full `claude -p` session that loads the user's entire MCP/skills context, so a user running the loop as documented burns a large share of their plan's usage limit on measurements that were never valid.

All four defects below were hit in one real session (Claude Code 2.1.247 desktop, Windows 11 Pro, Python 3.14.6, `anthropic-skills` plugin build) and confirmed present in this repo at current `main`.

## Defect 1 — `select.select()` on a subprocess pipe → every probe dies with WinError 10038

`run_eval.py:108`:

```python
ready, _, _ = select.select([process.stdout], [], [], 1.0)
```

On Windows, `select()` supports **sockets only**. Every single probe raises `[WinError 10038] An operation was attempted on something that is not a socket`, gets swallowed as `Warning: query failed`, and counts as "not triggered". The loop then prints a clean-looking `precision=100% recall=0%` table and proceeds to "improve" the description.

**Fix that worked locally:** replace the select loop with a daemon reader thread pumping `process.stdout.read1(8192)` into a `queue.Queue`, and `queue.get(timeout=1.0)` in the main loop. Cross-platform, keeps the early-kill-on-trigger behavior.

## Defect 2 — `find_project_root()` walks up to `$HOME` and writes probe files into the user's **global** `~/.claude/commands`

`run_eval.py:22` walks up from `Path.cwd()` until it finds a `.claude` directory. The documented invocation (`cd <skill-creator dir> && python -m scripts.run_loop ...`) starts from a plugin directory with no `.claude`, so on a default setup the walk reaches the user's home — which **does** contain `~/.claude` — and `run_eval.py:53` then writes every probe command file into the user's global `~/.claude/commands/`.

Consequences observed:
- Probe skills (`<name>-skill-<uuid>`) appeared in the available-skills list of **every other session on the machine**, including unrelated interactive ones.
[...]

PR #1694, opened 2026-08-31
fix(skill-creator): resolve Windows eval harness pipe reading and probe command leakage

AbhiPra24:
## Summary
- Replaced `select.select()` on pipes in `run_eval.py` with a thread-safe `queue.Queue` reader thread, avoiding WinSock-only `select()` crashes on Windows anonymous pipes.
- Updated `find_project_root()` to explicitly avoid matching `Path.home()` (`~/.claude`), ensuring probe command files are never written to the user's global command directory.
- Added `cleanup_probe_commands()` invoked at start and in `finally` blocks across `run_eval.py` and `run_loop.py` to guarantee probe files (`*-skill-*.md`) are safely deleted even when runs are interrupted or fail.
- Added explicit `utf-8` encoding across file operations in `run_eval.py`, `run_loop.py`, `improve_description.py`, and `utils.py` to avoid Windows ANSI/cp1252 decoding errors.

Fixes #1692

cc @98zc5g5jyw-arch for review

### DIFF
--- a/skills/skill-creator/scripts/improve_description.py
+++ b/skills/skill-creator/scripts/improve_description.py
@@ -186,7 +186,7 @@ Please respond with only the new description text in <new_description> tags, not
     if log_dir:
         log_dir.mkdir(parents=True, exist_ok=True)
         log_file = log_dir / f"improve_iter_{iteration or 'unknown'}.json"
-        log_file.write_text(json.dumps(transcript, indent=2))
+        log_file.write_text(json.dumps(transcript, indent=2), encoding="utf-8")
 
     return description
 
@@ -205,10 +205,10 @@ def main():
         print(f"Error: No SKILL.md found at {skill_path}", file=sys.stderr)
         sys.exit(1)
 
-    eval_results = json.loads(Path(args.eval_results).read_text())
+    eval_results = json.loads(Path(args.eval_results).read_text(encoding="utf-8"))
     history = []
     if args.history:
-        history = json.loads(Path(args.history).read_text())
+        history = json.loads(Path(args.history).read_text(encoding="utf-8"))
 
     name, _, content = parse_skill_md(skill_path)
     current_description = eval_results["description"]
--- a/skills/skill-creator/scripts/run_eval.py
+++ b/skills/skill-creator/scripts/run_eval.py
@@ -8,9 +8,10 @@ for a set of queries. Outputs results as JSON.
 import argparse
 import json
 import os
-import select
+import queue
 import subprocess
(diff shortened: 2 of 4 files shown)


Hahaknight (commented):
Verified on real Windows 11 (Python 3.13). This fixes defects 1, 2 and 4 from #1692 — but as written, **defect 3 (twin collision capping recall at ~1/N) appears to remain**, see last section.

**Defect 1 (select on pipes) — fixed, verified.** `main` still crashes in my environment: `select.select([p.stdout], ...)` raises `OSError: [WinError 10093]` (WSANOTINITIALISED — same class as the reporter's 10038: select() only supports sockets on Windows). This PR's thread-pump + `queue.Queue` pattern tested on the same platform: receives all subprocess output correctly. Cross-platform and keeps early-kill semantics.

**Defect 2 (global ~/.claude/commands leak) — fixed, verified** (3 cases):
- from a git-clone subdir with no `.claude`: returns the repo root (via the new `.git` marker), not `$HOME` ✅
- from a bare directory with no `.git`/`.claude` anywhere below home: returns cwd, never walks up into `$HOME` ✅
- `cleanup_probe_commands(root, "myproj")` removes only `myproj-skill-*.md` and leaves `other-skill-xyz.md` / `unrelated.md` untouched; the no-name variant sweeps all `*-skill-*.md` ✅ (good orphan cleanup for previously-interrupted runs)

**Defect 4 (encoding) — fixed.** `run_loop.py` eval-set read and all report/JSON writes now explicit `utf-8`, matching #1687/#1690.

**Defect 3 (parallel twins → recall ~1/N) — not fixed by this PR.** `run_single_query` still writes `{skill_name}-skill-{uuid}.md` into the **shared** `project_root/.claude/commands/`, so while `ProcessPoolExecutor` runs ~10 probes concurrently, every `claude -p` session sees ~10 near-identical twins and invokes an arbitrary one — only the probe owning that UUID counts a hit. That's exactly the 11% train / 8% test plateau (~1/10) the reporter measured, which persisted "regardless of description". The reporter's local fix (per-probe `tempfile.mkdtemp()` root + `rmtree` in finally) took the same description from ~11% to 60/60 — worth adopting here or in a follow-up; it would also make the defect-2 leak structurally impossible rather than mitigated.

Minor: the reporter's last suggestion (loud failure when a large share of probes error, instead of feeding `recall=0%` into the improver) isn't addressed — a one-line guard in `run_eval` raising if e.g. >50% of futures raised would stop quota burn on invalid measurements.

PR #1720, opened 2026-09-03
fix(mcp-builder): write evaluation report as UTF-8

### LINKED ISSUE #1669: mcp-builder: evaluation.py discards a completed run on Windows — write_text() without encoding hits cp1252 on the report's emoji
## Summary

`skills/mcp-builder/scripts/evaluation.py:366` writes the report with

```python
args.output.write_text(report)
```

`Path.write_text()` uses the **locale** encoding. On a Windows default install that is cp1252, and the report contains emoji — the ✅/❌ correctness markers in `TASK_TEMPLATE`. So on Windows the run completes, every task is asked, every tool call is made, **every token is paid for**, and the results are then discarded by a `UnicodeEncodeError` on the last line of the program.

Nothing is written. The results only ever existed in memory, so the entire run has to be repeated.

```
  File ".../scripts/evaluation.py", line 366, in main
    args.output.write_text(report)
  File ".../pathlib/_local.py", line 555, in write_text
    return PathBase.write_text(self, data, encoding, errors, newline)
  ...
  File ".../encodings/cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '❌' in position 602:
character maps to <undefined>
```

Position 602 is the first ❌ — so ironically the crash only happens once at least one task has been marked incorrect. A run where everything passed writes fine, which makes this look intermittent.

`PYTHONIOENCODING=utf-8` does **not** help; that governs stdio, not `pathlib`.

## Fix

```python
args.output.write_text(report, encoding="utf-8")
```

Worth doing for `parse_evaluation_file` too if evaluation XML may contain non-ASCII — `ET.parse` handles its own declaration, but any other `read_text`/`open` in the scripts has the same exposure. There is prior art for exactly this sweep in skill-creator (#1591, #1371, #1199); mcp-builder appears to have been missed. #1602 touches `evaluation.py` but does not change this line.

## Repro

Run any evaluation to completion on Windows with `-o report.md`, with at least one task scoring incorrect.

## Environment

- Windows 11, Python 3.13.14, `anthropic` 1.0.0, `mcp` 1.29.1
- Model: `claude-sonnet-5`
- Target: stdio MCP server exposing 66 tools

## Why this one stings

It is the last statement executed. Everything expensive has already happened — 10 tasks, ~24 tool calls, the full agent loop — and the failure mode is total loss of the output rather than a degraded report. A one-word fix.

CJstate:
Fixes #1669.

`Path.write_text()` defaults to the locale encoding. On Windows that is cp1252, and the report contains the ✅/❌ correctness markers from `TASK_TEMPLATE`. As soon as at least one task is marked incorrect, the run raises `UnicodeEncodeError` on the final `write_text()` call, and the entire paid-for evaluation result is discarded.

`PYTHONIOENCODING` does not help (it governs stdio, not pathlib).

Pass an explicit `encoding="utf-8"` so the report survives on any locale.

### DIFF
--- a/skills/mcp-builder/scripts/evaluation.py
+++ b/skills/mcp-builder/scripts/evaluation.py
@@ -363,7 +363,7 @@ Examples:
         report = await run_evaluation(args.eval_file, connection, args.model)
 
         if args.output:
-            args.output.write_text(report)
+            args.output.write_text(report, encoding="utf-8")
             print(f"\n✅ Report saved to {args.output}")
         else:
             print("\n" + report)


CJstate:
Added independent verification of the failure mode, plus a sweep for sibling exposure.

**Reproduced on a Windows machine with a non-UTF-8 locale (cp936):**

```
locale preferred encoding: cp936
write_text(report)              -> UnicodeEncodeError: 'gbk' codec can't encode character '✅'
write_text(report, encoding=..) -> ok, round-trips byte-for-byte
```

The report content is `TASK_TEMPLATE` populated at line 263 with `correct_indicator="✅" if result["score"] else "❌"`, so the crash needs at least one incorrect task — a fully-passing run writes fine, which is what makes it look intermittent (as @Hahaknight noted on the issue, it is in fact deterministic once any task fails).

**Sibling sweep in `skills/mcp-builder/scripts/`:** `evaluation.py:366` is the only locale-encoding exposure. `parse_evaluation_file` uses `ET.parse`, which honours the XML declaration's own encoding; the remaining `json.dumps` calls only produce strings and never touch the filesystem. So the one-line fix is complete for this skill — no need to widen it.

PR #1724, opened 2026-09-04
mcp-builder: update evaluation.py default model to claude-sonnet-5

### LINKED ISSUE #1716: mcp-builder: evaluation.py defaults to claude-3-7-sonnet-20250219, and the low-score advice points back at that same model
The evaluation harness in `mcp-builder` still defaults to `claude-3-7-sonnet-20250219`, and the troubleshooting section recommends that same snapshot as the *upgrade* path. Line numbers are against `main` at the time of writing.

### Where

`skills/mcp-builder/scripts/evaluation.py`

```python
223:    model: str = "claude-3-7-sonnet-20250219",
324:    parser.add_argument("-m", "--model", default="claude-3-7-sonnet-20250219", help="Claude model to use (default: claude-3-7-sonnet-20250219)")
```

`skills/mcp-builder/reference/evaluation.md`

```
488:  -m, --model           Claude model to use (default: claude-3-7-sonnet-20250219)
599:- Use a more capable model (e.g., `claude-3-7-sonnet-20250219`)
```

### Why it matters

The harness exists to tell an MCP author whether their tool descriptions and schemas are good enough for an agent to use. Scoring them against a model three generations old measures the wrong thing: descriptions that a current model reads fine can fail on 3.7, and descriptions that 3.7 needs spelled out are over-specified for the models people actually deploy against. Anyone who runs `evaluation.py` without `-m` gets that older baseline silently, because nothing in the output says which model produced the score.

Line 599 compounds it. It sits under troubleshooting for low scores and offers the pinned default as the more capable alternative, so a reader following the advice moves nowhere. When that snapshot is eventually retired, the same default turns into a hard failure for every invocation that omits `-m`.

### Suggested fix

1. Point both defaults in `evaluation.py` at a current Sonnet-class snapshot, and update the two lines in `reference/evaluation.md` to match.
2. Reword the troubleshooting bullet so it names a step up from the default rather than the default itself, or drop the model suggestion and keep the description-quality advice around it.
3. Consider printing the model id in the run header, so a score is always attributable to a baseline.

Happy to send a PR if the direction looks right.

ExpertVagabond:
## What

`skills/mcp-builder/scripts/evaluation.py` defaulted to `claude-3-7-sonnet-20250219` in both the `run_evaluation()` signature and the `-m/--model` argparse option, and `reference/evaluation.md` recommended that same snapshot as the "more capable" alternative under troubleshooting. This PR points both defaults at `claude-sonnet-5`, updates the usage text in `reference/evaluation.md` to match, and rewords the troubleshooting bullet to suggest a step up from the default (`-m claude-opus-5`) rather than the default itself. It also prints the model id in the run header so a score is always attributable to a baseline, and updates the `--help` example that passed `claude-3-5-sonnet-20241022` so it no longer points at a retired snapshot.

## Why

The harness exists to tell an MCP author whether their tool descriptions are good enough for a current agent to use, and scoring against a three-generation-old model measures the wrong thing. `claude-sonnet-5` is the id this repo's own `skills/claude-api/shared/model-migration.md` lists as the replacement for `claude-3-7-sonnet-20250219`, and model ids in this repo are bare aliases with no date suffix, so the default will not break when an old snapshot is retired.

Fixes #1716

## Verification

- `python3 -m py_compile skills/mcp-builder/scripts/evaluation.py` exits 0.
- `evaluation.py --help` (argparse, with `connections` stubbed) prints `-m, --model MODEL     Claude model to use (default: claude-sonnet-5)` and exits 0.
- `grep -rn "claude-3-7\|claude-3-5" skills/mcp-builder` returns no matches.
- Note: running `--help` directly with `mcp` 2.1.1 installed fails in `connections.py` (`streamablehttp_client` import). That reproduces identically on `main` and is unrelated to this change.

### DIFF
--- a/skills/mcp-builder/reference/evaluation.md
+++ b/skills/mcp-builder/reference/evaluation.md
@@ -485,7 +485,7 @@ positional arguments:
 optional arguments:
   -h, --help            Show help message
   -t, --transport       Transport type: stdio, sse, or http (default: stdio)
-  -m, --model           Claude model to use (default: claude-3-7-sonnet-20250219)
+  -m, --model           Claude model to use (default: claude-sonnet-5)
   -o, --output          Output file for report (default: print to stdout)
 
 stdio options:
@@ -596,7 +596,7 @@ If many evaluations fail:
 ### Timeout Issues
 
 If tasks are timing out:
-- Use a more capable model (e.g., `claude-3-7-sonnet-20250219`)
+- Use a more capable model than the default (e.g., `-m claude-opus-5`)
 - Check if tools are returning too much data
 - Verify pagination is working correctly
 - Consider simplifying complex questions
--- a/skills/mcp-builder/scripts/evaluation.py
+++ b/skills/mcp-builder/scripts/evaluation.py
@@ -220,10 +220,10 @@ TASK_TEMPLATE = """
 async def run_evaluation(
     eval_path: Path,
     connection: Any,
-    model: str = "claude-3-7-sonnet-20250219",
+    model: str = "claude-sonnet-5",
 ) -> str:
     """Run evaluation with MCP server tools."""
-    print("🚀 Starting Evaluation")
+    print(f"🚀 Starting Evaluation (model: {model})")
 
     client = Anthropic()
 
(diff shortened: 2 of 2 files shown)


Alsami1998:
love u

ExpertVagabond:
Bumping this since #1716 is still open and the diff is small.

What it does: evaluation.py and reference/evaluation.md both defaulted to claude-3-7-sonnet-20250219, so the harness scored MCP tool descriptions against a three-generation-old model, and the low-score troubleshooting advice pointed back at that same snapshot. This moves both defaults to claude-sonnet-5, repoints the troubleshooting bullet at claude-opus-5 so it suggests a step up rather than the default, prints the model id in the run header so a score is attributable to a baseline, and drops the retired claude-3-5-sonnet-20241022 from the --help example.

The ids are bare aliases with no date suffix, matching what this repo's own skills/claude-api/shared/model-migration.md lists as the replacement, so the default won't break on the next snapshot retirement.

Still merges clean, no conflicts. py_compile passes and --help shows the new default.

Happy to rebase or split it up if that makes review easier.

Alsami1998:
#1724 (comment)

PR #1742, opened 2026-09-08
fix(mcp-builder): support mcp>=2 streamable_http_client import and custom headers

### LINKED ISSUE #1668: mcp-builder: requirements.txt allows mcp>=2, where connections.py fails at import (streamablehttp_client renamed)
## Summary

`skills/mcp-builder/scripts/requirements.txt` pins `mcp>=1.1.0` with no upper bound. The MCP Python SDK is now at 2.x, where `streamablehttp_client` has been **renamed** to `streamable_http_client`. `connections.py` imports the old name at module scope, so a fresh install of the documented requirements fails at import — before any MCP server, evaluation file, or API key is involved.

This is the first thing anyone following the Phase-4 instructions hits today.

## Repro

```bash
python -m venv .venv
.venv/bin/pip install -r skills/mcp-builder/scripts/requirements.txt   # resolves mcp 2.x
.venv/bin/python skills/mcp-builder/scripts/evaluation.py -t stdio -c node -a server.js eval.xml
```

```
  File ".../scripts/evaluation.py", line 19, in <module>
    from connections import create_connection
  File ".../scripts/connections.py", line 10, in <module>
    from mcp.client.streamable_http import streamablehttp_client
ImportError: cannot import name 'streamablehttp_client' from 'mcp.client.streamable_http'
```

Confirmed on `mcp` 2.1.1:

```python
>>> import mcp.client.streamable_http as s
>>> [n for n in dir(s) if 'client' in n.lower()]
['ClientMessageMetadata', 'create_mcp_http_client', 'streamable_http_client']
```

## Fix

Either cap the requirement:

```
mcp>=1.1.0,<2
```

or support both names in `connections.py`:

```python
try:
    from mcp.client.streamable_http import streamable_http_client
except ImportError:  # mcp < 2
    from mcp.client.streamable_http import streamablehttp_client as streamable_http_client
```

The second is preferable — pinning `<2` in a shared environment holds back anything else on the machine using the SDK, which pushes users toward a throwaway venv for what should be a two-line install.

## Environment

- Windows 11, Python 3.13.14
- `mcp` 2.1.1 (fails), 1.29.1 (works)
- `anthropic` 1.0.0

## Context

Found while running the Phase-4 evaluation against a stdio MCP server exposing 66 tools. It was one of several independent failures between "follow the skill's instructions" and "get a score"; the others are #1390, #1386, and the report-write bug filed alongside this one. Happy to send a PR for whichever fix you prefer.

Kuldeeep18:
Fixes #1668

### Problem
In `mcp>=2.0.0`, `streamablehttp_client` was renamed to `streamable_http_client`, and custom HTTP headers are configured via `create_mcp_http_client` / `http_client` rather than as a direct kwarg to `streamable_http_client`.

Because `skills/mcp-builder/scripts/connections.py` unconditionally imported `streamablehttp_client` at module scope, importing `connections` or running `evaluation.py` fails on startup when installing `requirements.txt` (which specifies `mcp>=1.1.0` without an upper bound):

```
  File ".../scripts/evaluation.py", line 19, in <module>
    from connections import create_connection
  File ".../scripts/connections.py", line 10, in <module>
    from mcp.client.streamable_http import streamablehttp_client
ImportError: cannot import name 'streamablehttp_client' from 'mcp.client.streamable_http'
```

### Solution
1. **Compatible import:** In `skills/mcp-builder/scripts/connections.py`, import `streamable_http_client` and `create_mcp_http_client` with fallback to `streamablehttp_client as streamable_http_client` for `mcp<2.0.0`.
2. **Context configuration:** In `MCPConnectionHTTP._create_context()`:
   - When `create_mcp_http_client` is available, create `http_client` with custom headers and pass to `streamable_http_client(url=self.url, http_client=http_client)`.
   - On `mcp<2.0.0`, pass `headers=self.headers` directly.
3. **Regression tests:** Added `skills/mcp-builder/scripts/test_connections.py` testing import resolution, stdio/sse/http transport factories, context manager creation, and validation error cases.

### Verification
- Ran `pytest skills/mcp-builder/scripts/test_connections.py -v`: all 5 tests passed.
- Verified `python skills/mcp-builder/scripts/evaluation.py --help` under `mcp 2.1.1` and Python 3.14: cleanly displays full usage and exits 0.
- Verified compilation with `python -m py_compile`.

### DIFF
--- a/skills/mcp-builder/scripts/connections.py
+++ b/skills/mcp-builder/scripts/connections.py
@@ -7,7 +7,18 @@ from typing import Any
 from mcp import ClientSession, StdioServerParameters
 from mcp.client.sse import sse_client
 from mcp.client.stdio import stdio_client
-from mcp.client.streamable_http import streamablehttp_client
+
+try:
+    from mcp.client.streamable_http import streamable_http_client
+
+    try:
+        from mcp.client.streamable_http import create_mcp_http_client
+    except ImportError:
+        create_mcp_http_client = None
+except ImportError:
+    # Fallback for mcp < 2.0 where function was named streamablehttp_client
+    from mcp.client.streamable_http import streamablehttp_client as streamable_http_client
+    create_mcp_http_client = None
 
 
 class MCPConnection(ABC):
@@ -106,7 +117,14 @@ class MCPConnectionHTTP(MCPConnection):
         self.headers = headers or {}
 
     def _create_context(self):
-        return streamablehttp_client(url=self.url, headers=self.headers)
+        if create_mcp_http_client is not None:
+            http_client = create_mcp_http_client(headers=self.headers) if self.headers else None
+            return streamable_http_client(url=self.url, http_client=http_client)
+        else:
+            try:
+                return streamable_http_client(url=self.url, headers=self.headers)
+            except TypeError:
+                return streamable_http_client(url=self.url)
 
 
 def create_connection(
--- a/skills/mcp-builder/scripts/connections.py
+++ b/skills/mcp-builder/scripts/connections.py
@@ -119,6 +119,8 @@ class MCPConnectionHTTP(MCPConnection):
     def _create_context(self):
         if create_mcp_http_client is not None:
             http_client = create_mcp_http_client(headers=self.headers) if self.headers else None
+            if http_client is not None and self._stack is not None:
+                self._stack.push_async_callback(http_client.aclose)
             return streamable_http_client(url=self.url, http_client=http_client)
         else:
             try:
--- a/skills/mcp-builder/scripts/test_connections.py
+++ b/skills/mcp-builder/scripts/test_connections.py
@@ -1,0 +1,74 @@
+"""Regression tests for connections.py (anthropics/skills#1668).
+
+Verifies that:
+- connections.py imports cleanly under both mcp >= 2.x and mcp < 2.x
+- create_connection instantiates the correct connection class for stdio, sse, and http transports
+- argument validation operates as expected for missing parameters or unsupported transports
+"""
+
+import pytest
+import sys
+from pathlib import Path
+
+# Add scripts directory to sys.path
+SCRIPT_DIR = Path(__file__).resolve().parent
+if str(SCRIPT_DIR) not in sys.path:
+    sys.path.insert(0, str(SCRIPT_DIR))
+
+from connections import (
+    create_connection,
+    MCPConnectionStdio,
+    MCPConnectionSSE,
+    MCPConnectionHTTP,
+    streamable_http_client,
+)
+
+
+def test_streamable_http_client_imported():
+    """Verify that streamable_http_client function is successfully resolved."""
+    assert callable(streamable_http_client)
+
+
+def test_create_stdio_connection():
+    """Verify stdio connection creation."""
+    conn = create_connection("stdio", command="node", args=["server.js"])
+    assert isinstance(conn, MCPConnectionStdio)
+    assert conn.command == "node"
+    assert conn.args == ["server.js"]
+
+
+def test_create_sse_connection():
+    """Verify sse connection creation."""
+    conn = create_connection("sse", url="http://localhost:8000/sse")
+    assert isinstance(conn, MCPConnectionSSE)
+    assert conn.url == "http://localhost:8000/sse"
+
+
+def test_create_http_connection():
+    """Verify http and streamable_http transport variants."""
+    for transport in ["http", "streamable_http", "streamable-http", "HTTP"]:
+        conn = create_connection(transport, url="http://localhost:8000/mcp")
+        assert isinstance(conn, MCPConnectionHTTP)
+        assert conn.url == "http://localhost:8000/mcp"
+        # Verify context manager can be created
+        ctx = conn._create_context()
+        assert ctx is not None
+
+
+def test_validation_errors():
+    """Verify parameter validation."""
+    with pytest.raises(ValueError, match="Command is required"):
+        create_connection("stdio")
+
+    with pytest.raises(ValueError, match="URL is required"):
+        create_connection("sse")
+
+    with pytest.raises(ValueError, match="URL is required"):
+        create_connection("http")
+
+    with pytest.raises(ValueError, match="Unsupported transport type"):
+        create_connection("invalid_transport")
+
+
+if __name__ == "__main__":
+    pytest.main([__file__, "-v"])
--- a/skills/mcp-builder/scripts/test_connections.py
+++ b/skills/mcp-builder/scripts/test_connections.py
@@ -4,17 +4,22 @@ Verifies that:
 - connections.py imports cleanly under both mcp >= 2.x and mcp < 2.x
 - create_connection instantiates the correct connection class for stdio, sse, and http transports
 - argument validation operates as expected for missing parameters or unsupported transports
+- caller-owned custom HTTP client is properly closed on normal exit and on initialization failure
 """
 
-import pytest
-import sys
+from contextlib import asynccontextmanager
 from pathlib import Path
+import sys
+
+import anyio
+import pytest
 
 # Add scripts directory to sys.path
 SCRIPT_DIR = Path(__file__).resolve().parent
 if str(SCRIPT_DIR) not in sys.path:
     sys.path.insert(0, str(SCRIPT_DIR))
 
+import connections
 from connections import (
     create_connection,
     MCPConnectionStdio,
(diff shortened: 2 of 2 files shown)


yao23 (commented):
Static review of 8a2d825: the custom-header path creates an HTTP client but does not enter or close it. The [SDK only manages clients it creates internally](https://github.com/modelcontextprotocol/python-sdk/blob/v2.1.1/src/mcp/client/streamable_http.py#L664-L678); caller-supplied clients remain caller-owned.

The current HTTP test uses no headers and never enters the context, so it misses this path. Could we register the custom client with the connection’s AsyncExitStack and add async regressions covering both normal exit and initialization failure?

AI-assisted static review; I haven’t run a reproduction yet.

yao23:
Following up with an independent local reproduction at `8a2d825fa2d5fbaf2d375554985bd80503e4eda0`, using Python 3.11.15 with both `mcp==1.29.1` and `mcp==2.1.1`.

Both SDK versions produced the same results:

* Unmodified PR: **5 passed, 2 failed**. Normal exit and an explicitly asserted `session.initialize()` failure both left the caller-owned client open. The failures were at `client.is_closed is True`.
* With the minimal fix: **7 passed**. The original five tests also passed separately.

The fix registers cleanup immediately after creating the custom client:

```python
if http_client is not None:
    self._stack.push_async_callback(http_client.aclose)
```

This lets the transport exit before closing the caller-owned client and also covers initialization failure.

The regressions retain the real SDK transport lifecycle and the actual client created by `MCPConnectionHTTP`, stubbing only network I/O and `ClientSession` responses. They do not mock `is_closed` or close the client from a stub. This is an isolated lifecycle regression, not a live-network end-to-end test.

Implementation and validation were AI-assisted.

yao23:
I independently reproduced the custom-header HTTP client lifecycle issue at
`8a2d825fa2d5fbaf2d375554985bd80503e4eda0` on CPython 3.11.15 with both
`mcp==1.29.1` and `mcp==2.1.1`. Both isolated environments used
`anthropic==1.4.0`, `pytest==9.1.1`, and `pytest-asyncio==1.4.0`; imports and
`evaluation.py --help` passed in both.

The tests retain the real client created by `MCPConnectionHTTP` and the real SDK
`streamable_http_client` lifecycle. Only the network writer and `ClientSession`
response are stubbed. The stubs do not close the client, mock `is_closed`, or
fabricate closure. The initialization-error case first confirms the specific
expected exception, then checks the real client's state. This is therefore a
transport-isolated regression, not a live-network end-to-end test.

On the unmodified head, each SDK run produced `5 passed, 2 failed`; both failures
were only at the final `client.is_closed is True` assertion. Registering
`http_client.aclose` on the connection's `AsyncExitStack` immediately after
creation made both runs report `7 passed`. The five existing connection tests
also pass independently under both versions (`5 passed, 2 deselected`).

Re-run from a clean checkout, with the two patch files in the current directory:

```sh
git checkout --detach 8a2d825fa2d5fbaf2d375554985bd80503e4eda0
python3.11 -m venv .venv-mcp-1.29.1
python3.11 -m venv .venv-mcp-2.1.1
.venv-mcp-1.29.1/bin/python -m pip install 'mcp==1.29.1' 'anthropic==1.4.0' 'pytest==9.1.1' 'pytest-asyncio==1.4.0'
.venv-mcp-2.1.1/bin/python -m pip install 'mcp==2.1.1' 'anthropic==1.4.0' 'pytest==9.1.1' 'pytest-asyncio==1.4.0'
.venv-mcp-1.29.1/bin/python skills/mcp-builder/scripts/evaluation.py --help
.venv-mcp-2.1.1/bin/python skills/mcp-builder/scripts/evaluation.py --help
git apply regression-tests.patch
.venv-mcp-1.29.1/bin/python -m pytest -q skills/mcp-builder/scripts/test_connections.py
.venv-mcp-2.1.1/bin/python -m pytest -q skills/mcp-builder/scripts/test_connections.py
git apply minimal-fix.patch
.venv-mcp-1.29.1/bin/python -m pytest -q skills/mcp-builder/scripts/test_connections.py
.venv-mcp-2.1.1/bin/python -m pytest -q skills/mcp-builder/scripts/test_connections.py
```

<details>
<summary>Regression test patch</summary>

```diff
diff --git a/skills/mcp-builder/scripts/test_connections.py b/skills/mcp-builder/scripts/test_connections.py
index f914eba..3078e1b 100644
--- a/skills/mcp-builder/scripts/test_connections.py
+++ b/skills/mcp-builder/scripts/test_connections.py
@@ -6,10 +6,15 @@ Verifies that:
 - argument validation operates as expected for missing parameters or unsupported transports
 """
 
+from contextlib import asynccontextmanager
+
+import anyio
 import pytest
 import sys
 from pathlib import Path
 
+import mcp.client.streamable_http as streamable_http_module
+
 # Add scripts directory to sys.path
 SCRIPT_DIR = Path(__file__).resolve().parent
 if str(SCRIPT_DIR) not in sys.path:
@@ -24,6 +29,58 @@ from connections import (
 )
 
 
+class ExpectedInitializationError(RuntimeError):
+    """Sentinel error used to verify initialization-failure cleanup."""
+
+
+class StubClientSession:
+    """Minimal session stub; transport and HTTP-client lifecycles stay real."""
+
+    initialize_error = None
+
+    def __init__(self, read, write):
+        self.read = read
+        self.write = write
+
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        return False
+
+    async def initialize(self):
+        if self.initialize_error is not None:
+            raise self.initialize_error
+
+
+@pytest.fixture
+def captured_production_http_client(monkeypatch):
+    """Capture the real client passed into the real SDK transport context."""
+    captured = {}
+    real_streamable_http_client = streamable_http_client
+
+    @asynccontextmanager

**Comment by Kuldeeep18** [comment]
Thanks for the thorough review and independent reproduction, @yao23!

You're spot on—caller-supplied HTTP clients remain caller-owned in the MCP SDK, so failing to register cleanup left the custom client open.

In commit 0ba9e31, I have updated the PR to:
1. Register `http_client.aclose` via `self._stack.push_async_callback(http_client.aclose)` when custom headers create an HTTP client.
2. Add isolated lifecycle regression tests covering both normal context exit and initialization failure without requiring live network calls.

All 7 tests are passing. Thanks again for catching this!

**Comment by yao23** [comment]
> Kuldeeep18

Thanks for incorporating this so quickly!

One small test-hardening suggestion: could the lifecycle tests assert that the custom client was actually captured before checking its state?

```python
client = captured_production_http_client.get("client")
assert client is not None
assert client.is_closed is True
```

With the current conditional checks, a regression that stops creating or passing the custom client could make the tests pass vacuously. This should preserve the network-isolated setup while ensuring the intended custom-header path was exercised.

AI-assisted follow-up review.

Kuldeeep18:
Good catch, @yao23! Updated in b26eba8:

1. Added explicit \ssert client is not None\ in both lifecycle tests (\	est_http_custom_headers_client_closed_on_exit\ and \	est_http_custom_headers_client_closed_on_init_failure\) to ensure the custom client was actually captured and that the tests cannot pass vacuously.
2. Added \@pytest.mark.skipif(connections.create_mcp_http_client is None, ...)\ so that environments running legacy \mcp < 2\ (where \streamable_http_client\ doesn't use caller-owned custom HTTP clients) cleanly skip these lifecycle tests rather than failing.
3. Rebased on latest \upstream/main\.

All 7 tests pass cleanly. Thanks again for the follow-up review!

98zc5g5jyw-arch:
Independent verification: checked out the PR branch and ran the new suite on a clean venv (Python 3.11, latest `mcp` 2.x, pytest 9.1): **7/7 passed**, including the caller-owned HTTP client being closed on both normal exit and initialization failure. The mcp<2 fallback path is guarded by `skipif(create_mcp_http_client is None)` and degrades gracefully. Import shim is correct for both API generations. Minor observation (non-blocking): in the `mcp>=2` branch, if `self._stack` were None the created `http_client` would not be registered for cleanup — unlikely to occur in practice, but a comment would help future readers. LGTM for merge.

yao23 (approved):
Reviewed the current head `b26eba8b7fab86fa74de9bc1d76a3ca0a4f53bf9`. The caller-owned HTTP client is now registered with the connection’s `AsyncExitStack`, and the lifecycle regressions cover both normal context exit and initialization failure. The explicit `client is not None` assertions ensure the custom-header path cannot pass vacuously, while the legacy SDK path is guarded separately.

I previously reproduced the lifecycle defect at `8a2d825fa2d5fbaf2d375554985bd80503e4eda0` and verified the minimal correction under `mcp==1.29.1` and `mcp==2.1.1`. I have not rerun the rebased current head in this review, but I found no remaining blocking issue in the current diff.

Approved. AI-assisted review.

Kuldeeep18:
Hi maintainers, just a gentle check-in on this PR. It has been independently verified and approved by community reviewers with all regression tests passing, and the branch is clean and up to date with \main\. Whenever you have a moment, could someone from the team take a look for merge? Thank you!

PR #1755, opened 2026-09-11
fix(skill-creator): make the trigger-eval harness measure skill triggering

giovannibeltrame:
`skills/skill-creator/scripts/run_eval.py` and `run_loop.py` cannot currently measure whether a description triggers a skill. Every positive query scores 0% recall, and the same query alternates between 0.0 and 1.0 across runs — so `run_loop.py` optimizes descriptions against noise.

Five defects, all the same class: the experiment measures something other than the description.

## The defects

**The probe competes with the real skill.** `find_project_root()` walks up from cwd to the nearest `.claude/`, which lands in a tree that already registers the skill under test — the common case, since you run the eval from the repo holding the skill. Claude invokes the real name while the probe watches for its uuid-suffixed decoy, so every genuine trigger scores as a miss. That is the 100% precision / 0% recall signature. Each probe now builds and tears down its own root.

**The detector gives up at the first orienting command.** It returns `False` the moment a tool call is not `Skill` or `Read`. Reading the raw stream, Claude's opening move is usually a `Bash` locating itself (`pwd`, `ls`), and it consults the skill a turn later. That ordering is close to a coin flip, and it is the source of the non-determinism. A trigger now counts anywhere in the run, bounded by `MAX_PROBE_TURNS = 4` so a run that was never going to trigger stops before executing the whole queried task.

**Probes share a directory.** With more than one worker, several decoy commands are live at once. They differ only by uuid, so a worker can watch its own copy while Claude invokes another's.

**No stdin redirect.** `claude -p` is spawned without `stdin`, so every probe waits 3s for piped input that never arrives.

**An installed copy still wins.** A private probe root isolates project scope, but a user-level `~/.claude/skills/<name>` entry or a plugin reaches the probe anyway, and the real name outscores the decoy — the same false-miss, reported silently as 0%. This is easy to hit: testing a skill you actually use. The `system/init` event names what is registered, so the run now aborts with an error naming the offending copy instead of reporting a plausible zero.

## User-facing changes

`--project-root` becomes `--fixture`, reflecting what it now means: a *fixture copied into* each probe's sandbox. The old name invited pointing it at the real project — precisely the false-negative case the isolation exists to prevent. Skills whose queries only make sense somewhere specific need one: in a bare directory Claude answers "there's nothing here" rather than reaching for the skill, and every positive looks like a miss. Any `.claude/skills` or `.claude/commands` the fixture carries is stripped from the copy, so the decoy stays the only registered skill; the rest of the fixture survives intact, and the original is never modified. The default remains a bare temp dir.

`--timeout` rises from 30s to 120s. With the turn cap in place it is a safety net rather than the expected duration of a negative.

As a side effect, the harness stops writing probe files into the caller's `.claude/commands/`. That path is not gitignored, so an interrupted run left untracked files staged for an accidental commit.

## Verification

Against `skills/mcp-builder` (not installed locally, so the collision check passes), two positives and two negatives, 2 runs per query, 4 workers:

| Query | Expected | Stock | This PR |
|---|---|---|---|
| write a FastMCP server that wraps the Stripe API | triggers | 0.5 / 0.5 / 0.5 | 1.0 / 1.0 / 1.0 |
| I need to expose our internal billing API to Claude as tools it can call | triggers | 0.0 / 0.0 / 0.0 | 1.0 / 1.0 / 1.0 |
| the build/ directory has 2GB of stale artifacts, clean it up | no trigger | 0.0 / 0.0 / 0.0 | 0.0 / 0.0 / 0.0 |
| rename the getUser function to fetchUser across the repo | no trigger | 0.0 / 0.0 / 0.0 | 0.0 / 0.0 / 0.0 |

Three independent runs of each. The stock harness never scores a positive above 0.5 and reports one of them as a hard failure; this PR separates positives from negatives cleanly and repeats it.

Also checked: the collision guard fires against a skill installed at user scope (`skill-creator` itself) and exits with a message naming it; no temp roots survive a run; the fixture is never modified; both CLIs exit with a clear error on a missing `--fixture`; `quick_validate.py` passes.

Run 8 probes in ~20s, against minutes before — the turn cap, not a change in what is measured.

## Notes

This was found and fixed downstream while optimizing a skill description, against a vendored copy of `skill-creator` byte-identical to this repo's current `main`. The fifth defect above surfaced only when re-verifying the port here, against a skill installed at user scope. No behavior outside `skill-creator`'s eval scripts changes.

### DIFF
--- a/skills/skill-creator/SKILL.md
+++ b/skills/skill-creator/SKILL.md
@@ -389,6 +389,10 @@ python -m scripts.run_loop \
 
 Use the model ID from your system prompt (the one powering the current session) so the triggering test matches what the user actually experiences.
 
+Each probe runs in its own throwaway directory, so the probe never competes with a real installed copy of the skill and parallel probes can't see each other's commands. For a skill whose queries only make sense somewhere specific (a git repo, a project tree), a bare directory makes Claude answer "there's nothing here" instead of reaching for the skill, and every positive looks like a miss. Pass `--fixture <dir>` to copy a prepared fixture in as each probe's sandbox; probes mutate their copies, never the original. Build the smallest fixture that makes the queries plausible (for a git skill, a `git init` with a few branches is enough).
+
+The isolation covers project scope only. If the skill under test is also installed at user scope (`~/.claude/skills/`) or ships in a plugin, that copy answers the same queries under its real name and outscores the probe's decoy, so the run aborts with an error instead of reporting a fake 0%. Move or rename the installed copy for the duration of the eval.
+
 While it runs, periodically tail the output to give the user updates on which iteration it's on and what the scores look like.
 
 This handles the full optimization loop automatically. It splits the eval set into 60% train and 40% held-out test, evaluates the current description (running each query 3 times to get a reliable trigger rate), then calls Claude to propose improvements based on what failed. It re-evaluates each new description on both train and test, iterating up to 5 times. When it's done, it opens an HTML report in the browser showing the results per iteration and returns JSON with `best_description` — selected by test score rather than train score to avoid overfitting.
--- a/skills/skill-creator/scripts/run_eval.py
+++ b/skills/skill-creator/scripts/run_eval.py
@@ -9,27 +9,87 @@ import argparse
 import json
 import os
 import select
+import shutil
 import subprocess
 import sys
+import tempfile
 import time
 import uuid
 from concurrent.futures import ProcessPoolExecutor, as_completed
+from contextlib import contextmanager
 from pathlib import Path
 
 from scripts.utils import parse_skill_md
 
+DEFAULT_TIMEOUT = 120
 
-def find_project_root() -> Path:
-    """Find the project root by walking up from cwd looking for .claude/.
+# Triggering is an opening move: Claude orients itself (a pwd, an ls) for a
+# turn or two, then consults the skill. A probe that hasn't reached for it
+# within this many assistant turns is a miss; letting it continue means every
+# non-triggering run executes the entire queried task.
+MAX_PROBE_TURNS = 4
 
-    Mimics how Claude Code discovers its project root, so the command file
-    we create ends up where claude -p will look for it.
+class SkillAlreadyRegistered(RuntimeError):
+    """The skill under test is registered outside the probe root.
+
+    The probe registers a uuid-suffixed decoy and counts invocations of that
+    exact name. An installed copy carrying the same description answers the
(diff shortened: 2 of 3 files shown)


98zc5g5jyw-arch:
Thanks for this — the five defects are real and the verification table is convincing. One safety edge case in the fixture-stripping logic (`scripts/run_eval.py:84-88`) worth tightening before this lands:

**If the fixture contains `.claude` as a symlink** (a common dotfiles pattern), the strip logic deletes the symlink's *target* instead of the copy's entry:

- `copytree(fixture, root, symlinks=True, dirs_exist_ok=True)` copies the link as a link, so `root/.claude` is still a symlink.
- `leaked.is_symlink()` is then `False` (only the final component is checked), and `leaked.is_dir()` resolves *through* the intermediate link to the real directory.
- `shutil.rmtree(leaked)` therefore deletes the fixture-linked directory's `skills/` and `commands/` outside the probe root — which breaks the guarantee in this PR's description that the original is never modified.

Verified empirically on Python 3.9.6 (macOS default): with `fixture/.claude -> victim`, both `victim/skills` and `victim/commands` are removed. (Newer Pythons whose rmtree uses O_NOFOLLOW fail differently, with an opaque mid-probe error — a defect either way.)

Suggested fix: guard `.claude` itself before descending:

```python
claude = root / ".claude"
if claude.is_symlink() or claude.is_file():
    claude.unlink()
elif claude.is_dir():
    for leaked in (claude / "skills", claude / "commands"):
        if leaked.is_symlink() or leaked.is_file():
            leaked.unlink()
        elif leaked.is_dir():
            shutil.rmtree(leaked)
```

Everything else reviewed clean — probe isolation, the turn cap, stdin fix, and the init-event collision guard all look right, and `py_compile` passes on both scripts.

98zc5g5jyw-arch:
Reviewed the eval-harness isolation redesign — solid overall. Per-probe temp roots, the decoy-alone assertion on the `init` event (fail loudly instead of scoring a silent 0%), early-exit on first Skill/Read trigger, MAX_PROBE_TURNS, and `stdin=DEVNULL` all address real failure modes of the previous harness.

One minor fidelity note: `shutil.copytree(fixture, root, symlinks=True)` keeps symlinks inside the fixture as symlinks, so a probe that writes *through* a symlinked path would mutate the original fixture — a slight gap against the "probes mutate their copies, never the original" contract in the docstring. Consider `follow_symlinks` handling or documenting the exception. Non-blocking.

PR #1757, opened 2026-09-12
fix(skill-creator): scan full turn for skill triggers instead of bailing on first non-Skill tool (#1721)

### LINKED ISSUE #1721: skill-creator: trigger detection reports 0% recall for every skill
## Summary

## Cause

```python
    pending_tool_name = tool_name
    accumulated_json = ""
else:
    return False
```

```python
    if pending_tool_name:
```

```python
    ...
    return triggered
```

## Evidence

```
```

`Skill` call.

| | shipped | patched |
|---|---|---|

## Fix

## Also, possibly separate

## Environment

toprakeker:
Closes #1721

### Summary
Fixes a false-negative trigger detection defect where `run_single_query()` in `skill-creator`'s eval harness reported 0% recall for valid skills whenever an orientation tool (`Bash`, `Glob`, `Grep`, etc.) was invoked before the skill.

### Changes
1. **Full Turn Tool Scanning:** Extracted incremental event processing into `TriggerDetector`. Non-Skill/Read tools no longer trigger early `return False`; instead, detection continues until the skill is invoked or the turn completes (`result` event).
2. **Proper Block Lifecycle:** `content_block_stop` only checks accumulated input for the current block and does not prematurely finalize the entire turn as failed.
3. **Assistant Message Fallback:** Corrected the loop over `message['content']` so all content items are checked, rather than bailing on the first non-matching item.
4. **Unit Tests:** Added 14 unit tests in `test_run_eval_trigger.py` covering all multi-tool sequences, partial deltas, early orientation, and fallback scenarios.

Tested locally: 14/14 tests pass.

### DIFF
--- a/skills/skill-creator/scripts/run_eval.py
+++ b/skills/skill-creator/scripts/run_eval.py
@@ -32,6 +32,88 @@ def find_project_root() -> Path:
     return current
 
 
+class TriggerDetector:
+    """Incrementally consumes parsed `claude -p --output-format stream-json`
+    events for a single query and determines whether the target skill was
+    triggered.
+
+    Feed events one at a time via `process_event`. It returns True/False once
+    the outcome is certain (skill tool_use seen, or the turn/stream completed
+    without one), or None while the outcome is still undetermined. Any tool
+    use that is not the skill (e.g. Bash, Glob, Grep used to inspect files
+    first) must NOT short-circuit detection to False -- the model may still
+    invoke the skill later in the same turn.
+    """
+
+    def __init__(self, clean_name: str):
+        self.clean_name = clean_name
+        self.triggered = False
+        self._pending_tool_name = None
+        self._accumulated_json = ""
+
+    def _check_tool_use(self, tool_name: str, tool_input: dict) -> bool:
+        if tool_name == "Skill" and self.clean_name in tool_input.get("skill", ""):
+            return True
+        if tool_name == "Read" and self.clean_name in tool_input.get("file_path", ""):
+            return True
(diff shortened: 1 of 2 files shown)


98zc5g5jyw-arch:
Independent review of this fix for the trigger-eval false negatives in #1721:

**Verified locally**
- `python3 -m unittest scripts.test_run_eval_trigger` on this branch: 14/14 pass.
- Same suite against current `main`: fails (TriggerDetector absent), and `main` still carries the `return False`-on-first-non-Skill-tool bug (run_eval.py ~L141/153/168). So the defect is real, unfixed upstream, and the regression tests exercise exactly those paths.

**Code quality: good.** The `TriggerDetector` extraction is clean: non-Skill tools no longer short-circuit detection, `content_block_stop`/`message_stop` no longer finalize the turn prematurely, and the assistant-message fallback now scans the whole message instead of returning after the first item. Cosmetic nit only: `result_event(triggered: bool = False)` takes an unused parameter.

**⚠️ Conflict heads-up for maintainers:** open PR #1755 ("make the trigger-eval harness measure skill triggering") fixes the same three detection bugs in the same region of `run_eval.py`, plus probe isolation/fixture support and a per-run turn cap. These two PRs will collide — whichever merges first forces a rebase of the other. Worth coordinating whether to land #1755's isolation work together with this PR's detector extraction + unit tests. For reference, #1735 was an earlier attempt at the same fix and was closed unmerged.

*Advisory note from a PR-monitoring bot, not a maintainer decision.*

PR #1762, opened 2026-09-14
fix(skill-creator): preserve recorded token counts in benchmarks

rioyu123:
## Summary

Benchmark aggregation can report output character counts as tokens even when a run has an actual token count in `timing.json`. This happens when `grading.json` already contains a nonzero duration: the aggregator skips the timing file entirely.

This change reads recorded token counts independently of the duration fallback. The duration from grading still takes precedence, and missing or zero token counts keep the existing fallback behavior. Unreadable or malformed timing files also fall back rather than aborting aggregation now that the file is read for runs with an existing duration.

## Testing

- Added CLI regression tests for current and legacy layouts, summary and comparison output, duration precedence, and missing or invalid timing data.
- Confirmed the tests reproduce the incorrect token counts and comparison before the fix.
- From the repository root, ran `python -m unittest discover -s tests -p test_aggregate_benchmark.py -v` (11 passed on Windows and Linux).
- The tests also check that aggregation does not rewrite its input files.
- Tests live in the top-level `tests/` directory so they are not bundled into the packaged skill.


98zc5g5jyw-arch:
Reviewed and verified locally: fetched the PR branch and ran `tests/test_aggregate_benchmark.py` — 11/11 pass.

The fix correctly reads `timing.json` even when a duration is already recorded in `grading.json`, and the fallbacks behave sanely (missing/zero/unparseable token counts keep the `output_chars` fallback; non-dict timing and unreadable files are handled). Nice coverage including the legacy layout. 👍

PR #1764, opened 2026-09-14
fix(docx): detect orphaned comments

### LINKED ISSUE #1733: docx validate.py: comment present in comments.xml but anchored to nothing passes as valid

00200200:
## Summary

- detect comments that remain in `comments.xml` without any anchor in `document.xml`
- report the orphaned comment id alongside the existing marker validation errors

Fixes #1733

## Validation

I exercised `DOCXSchemaValidator.validate_comment_markers()` against a temporary OOXML directory containing an unanchored `<w:comment w:id="7">`; it now fails with:

```
comments.xml: comment id="7" is not anchored anywhere in document.xml
```

The existing marker-to-comment checks remain unchanged. `git diff --check` passes.

## AI assistance disclosure

This PR was prepared with assistance from OpenAI Codex. I reviewed the implementation and validation result and understand the change.


98zc5g5jyw-arch:
Heads-up: this change (orphaned-comment detection in `docx.py`) duplicates the same hunk already in open PR #1734 ("Detect orphaned docx comments" by rohitjain25), which additionally fixes the `//w:comment` → `./w:comment` XPath to only match direct children.

Also, your PRs #1764 → #1765 → #1766 appear stacked (each includes the previous one's hunks). If #1764 or #1765 merges first, the others will conflict or partially no-op. Suggest:
- closing this one in favor of #1734 (it's more complete), and
- rebasing #1765 / #1766 so each PR carries only its own change.

Thanks!

00200200:
Closing this duplicate in favor of #1734, which already contains the orphaned-comment fix and the related XPath correction. I will keep the UTF-8 redlining and trigger-evaluation changes in separate rebased PRs.

PR #1766, opened 2026-09-14
fix(skill-creator): keep scanning trigger events

### LINKED ISSUE #1721: skill-creator: trigger detection reports 0% recall for every skill

00200200:
## Summary

- keep scanning stream events when another tool appears before `Skill` or `Read`
- continue after an unrelated tool block instead of treating it as a failed trigger
- inspect every tool use in a fallback assistant message before deciding

Fixes #1721

## Validation

I ran `python3 -m py_compile` and exercised `run_single_query()` with a fake Claude CLI covering:

- another tool before the matching `Skill` call
- an unrelated `Skill` block followed by the matching one
- a fallback assistant message containing multiple tool uses

All three cases detected the skill trigger. `git diff --check` passes.

## AI assistance disclosure

This PR was prepared with assistance from OpenAI Codex. I reviewed the implementation and validation results and understand the change.


98zc5g5jyw-arch:
Flagging a collision risk before this grows further: `skills/skill-creator/scripts/run_eval.py` trigger detection is being edited by four open PRs from the same base (`e58c70bea3`):

- #1757 (toprakeker) — rewrites this exact logic as a `TriggerDetector` state machine for #1721
- #1755 (giovannibeltrame) — reworks the trigger-eval harness
- #1298 (MartinCajiao) — broader fix for the 0% recall issue, still actively updated
- this PR — minimal patch for the same bail-on-first-non-Skill-tool bug

Your patch at L138+ fixes the same short-circuit behavior #1757 addresses with a more complete state machine, and the hunks will conflict. Suggest coordinating with #1757's author (either fold this into it or rebase on top once one of them merges). Thanks!

00200200:
Closing this smaller overlapping patch in favor of #1757, which addresses the same trigger-scanning defect with the complete state-machine implementation and dedicated tests. Thanks for pointing out the collision.

PR #1763, opened 2026-09-14
Add resume-screening skill

maweis1981:
## Summary

Adds `skills/resume-screening`, a skill that turns job-application emails into an HR-ready screening report, and registers it as a plugin in `.claude-plugin/marketplace.json`.

It demonstrates a multi-step local workflow that combines scripts with Claude's judgment:

1. **Export**: `scripts/export_mail.applescript` pulls applications for a role from macOS Mail (subject keywords + date range), including resume attachments.
2. **Extract**: `scripts/extract.py` (stdlib only) parses candidate metadata and phone/email, strips recruiting-platform watermarks from PDF text, and de-duplicates repeat applicants.
3. **Evaluate**: Claude reads every resume in full and tiers candidates A/B/C using the rubric and red-flag checklist in `SKILL.md` (timeline overlaps, copy-paste artifacts, buzzword stuffing, inflated metrics, direction/salary mismatch), then writes `evaluation.json`.
4. **Report**: `scripts/build_report.py` renders a PDF via headless Chrome (summary, contact table with page references, per-candidate cards, interview questions) and appends the original shortlisted resumes with poppler's `pdfunite`.

Privacy is built into the instructions: all processing stays local, and the skill tells Claude not to upload or publish reports that contain candidate contact data unless explicitly asked.

## Notes

- Platform: macOS (Mail.app, Google Chrome, `brew install poppler`).
- Skill instructions are written in Chinese; it targets China-based recruiting flows (e.g. BOSS Zhipin forwards).
- License: MIT (`skills/resume-screening/LICENSE.txt`).
- Also maintained at https://github.com/maweis1981/claude-skills.

## Testing

Ran end-to-end on a real batch of 20 applications for an AI engineer role: export → extraction (all phones/emails recovered) → evaluation → a 48-page PDF with 11 resumes attached; verified rendering and page references. Fixed a headless-Chrome issue where the process keeps running after writing the PDF. No candidate data is included in this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01AkrHSBuySQ1WuDd2eEjgS2

PR #1703, opened 2026-09-01
Add md2video-audio skill

70v-Yoyo:
## Summary

Adds `md2video-audio`, a zero-cost skill that directly compiles Markdown documents into professional-grade MP4 videos with realistic human-like voiceovers.

* `skills/md2video-audio/SKILL.md` — Defines the workflow: Markdown documents are converted via Marp into presentation slides and paired with Edge-TTS to generate finished MP4 videos.
* `skills/md2video-audio/LICENSE` — Open-source license.
* `.claude-plugin/marketplace.json` — Registers the skill as a plugin entry.
* Local video rendering with online text-to-speech generation through `edge-tts`, no OpenAI or Alibaba Cloud dependencies, and no subscription costs required.

## Testing

Tested locally across tech tutorials, reading notes, project reviews, and data analysis documents to ensure stable one-click compilation from text to ready-to-upload MP4 videos for platforms like TikTok, Bilibili, and YouTube.

PR #525, opened 2026-03-05
Add pyxel skill for retro game development

kitao:
Add a Pyxel skill for creating, debugging, and verifying retro games in Python. The skill guides implementation, headless input-driven runs, direct frame inspection, and task-specific state checks. Separate references cover Pyxel behavior, presentation, and broader evidence for requested release checks and audits.

This contribution registers `./skills/pyxel` in the existing `example-skills` marketplace plugin and includes the skill, three reference documents, and an MIT `LICENSE.txt`. The contents match [pyxel-skill 1.4.2](https://github.com/kitao/pyxel-skill/releases/tag/v1.4.2), maintained in [kitao/pyxel-skill](https://github.com/kitao/pyxel-skill).

[pyxel-mcp](https://github.com/kitao/pyxel-mcp) supplies the eight observation tools and requires Python 3.11+ and Pyxel 2.9.6+. Configure it separately with the stdio command `uvx --from 'pyxel-mcp>=1.3.1' pyxel-mcp`. Appending `install` prints client-specific setup examples; apply the registration with this versioned command and restart the client. Version 1.3.1 corrects the reported cycling of audio tone, volume, and effect sequences. The skill describes a limited fallback when the server is unavailable and requires reporting verification limitations.

Validation:

- Agent Skills reference validator, this repository's `quick_validate.py`, and Claude Code marketplace validation passed.
- Isolated installation of the `example-skills` plugin passed. All five installed Pyxel files match the submitted contents; references and marketplace registration resolve. The temporary test marketplace used a distinct name because the official Anthropic name is reserved for its GitHub source.
- The installed skill's documented command connected to published pyxel-mcp 1.3.1, exposed all eight tools, and executed its audio example with the corrected metadata.
- Canonical pyxel-skill tests passed on Python 3.11, 3.12, and 3.14 (six tests each); [Linux CI passed](https://github.com/kitao/pyxel-skill/actions/runs/35685421280).
- The canonical launcher passed 29 integration checks with Pyxel 2.9.9 and 30 with the minimum supported Pyxel 2.9.6, covering all eight tools, input, deterministic state and images, output boundaries, failure paths, and audio rendering. Representative play and win frames were inspected directly; audio quality was not judged by listening.
- Canonical document and license matching, local links, and `git diff --check` passed.

PR #1771, opened 2026-09-15
feat(skills): add proofcore-contract-auditor for smart contract notarization

ProofCore-Protocol:
### Summary
Adds `proofcore-contract-auditor`, an Agent Skill for Web3 developers that performs automated static analysis of Solidity and Rust smart contracts and anchors cryptographic audit proofs onto the public TON Blockchain using ProofCore's zero-storage Merkle protocol.

### What this skill does
- **Security Audit:** Scans `.sol` and `.rs` files for common vulnerabilities (reentrancy, overflow, access control).
- **Data Boundary Isolation:** Explicitly isolates untrusted contract source code as passive data to prevent indirect prompt injection.
- **Cryptographic Notarization:** Generates an immutable SHA-256 Merkle root of the audit report and anchors it on-chain without storing proprietary contract code on remote servers.
- **Safe Execution & Fallback:** Binds natively to `mcp__proofcore__seal_content` if an MCP server is configured, with a safe file-based Python fallback (`scripts/anchor.py`) that prevents shell command injection.

### Verification & Testing
- Validated against the `agentskills.io` standard format.
- Tested locally with Claude Code CLI.
- No root or elevated privileges required.


98zc5g5jyw-arch:
Reviewed the new skill as a fresh pass (head `1b03fa4`) — thanks for the clear write-up. The mechanical checks pass, but the service integration needs resolving before this fits this repo.

**Checks that pass**
- ✅ Layout & frontmatter: `skills/proofcore-contract-auditor/`, kebab-case `name` matching the directory, 241-char description — `quick_validate.py` → "Skill is valid!".
- ✅ `scripts/anchor.py` compiles, is stdlib-only, does no shell interpolation (a single `urlopen` call), and sets a request timeout.
- ✅ Scope: only the two new files; nothing else touched.
- ✅ The "treat `*.sol` / `*.rs` as passive data" isolation note is the right instinct.

**To address**
1. **The notarization step is load-bearing and uploads by default.** Phase 2 says to anchor "as part of the requested audit workflow", i.e. every audit is expected to send the report off-machine; `anchor.py` then POSTs the report *content* (`payload.content`, not just a digest) to `api.proofcore.org`, uploading whichever path it is given. The "zero-storage" property isn't verifiable client-side, and the Tertiary Route has users wire in ProofCore's MCP server. The rest of `./skills` stays local-first — I couldn't find another skill that requires a specific external commercial service to complete its core workflow. Suggestion: make anchoring explicitly opt-in (compute the SHA-256 locally, show it, upload only after an explicit user confirmation) and state plainly in the skill what leaves the machine; or ship ProofCore as your own plugin/marketplace entry rather than inside `anthropics/skills`.
2. **The `npx` target isn't published yet.** `npx @proofcore/mcp-server` currently 404s on the npm registry (checked registry.npmjs.org and unpkg today), so the third route can't work — and instructions in a widely-used repo pointing `npx` at an unpublished scoped package are a dependency-confusion hazard. Publish it, or drop that snippet, before landing.
3. **Remote content is appended to the model's output.** Phase 3 instructs appending the server-returned `citation_markdown` directly to the final report — please treat that response as untrusted third-party data (sanitize it, or restrict to the returned URL).
4. **Missing marketplace registration.** Other new-skill PRs add a `.claude-plugin/marketplace.json` entry (#1763, #1703); without one this skill isn't installable from the marketplace.
5. **The audit half is thin.** "Run standard static analysis (e.g., Slither … Cargo …)" gives no install steps, no output parsing, no severity rules — yet `allowed-tools` pre-approves a broad `Bash(slither *)` / `Bash(cargo clippy *)` / `Bash(python3 *)` set (and is experimental in the spec, so support varies by client). Spell out the concrete commands and how findings map to the [CRITICAL…LOW] buckets — or frame the skill as notarization-first if that's the real intent.
6. Nits: 18 of the 19 skills here also ship a `LICENSE.txt` (this one carries frontmatter only); `${CLAUDE_SKILL_DIR}` is a documented Claude Code substitution so it works there, though no other skill in this repo relies on it; and the flow doesn't say what directory `audit_report.md` lives in between Phase 1 and Phase 2.

Happy to re-review once adjusted. (This repo has no CI workflows, so there's nothing CI-side to wait on.) Checked via static review + `quick_validate.py` + `py_compile`; the upload script was intentionally not executed.

ProofCore-Protocol:
Thanks for the thorough review and constructive feedback! All 6 points have been addressed in the latest commit:

1. **Local-First & Opt-In Anchoring:** Phase 2 is now explicitly opt-in. Furthermore, `anchor.py` has been refactored to compute the SHA-256 hash strictly client-side and transmits only the digest commitment envelope (`{"audit_report_sha256": "..."}`). Raw contract code or full reports never leave the machine.
2. **Removed Unpublished NPX Snippet:** Removed the `npx` snippet entirely to prevent dependency-confusion hazards. Fallback relies cleanly on the bundled stdlib-only Python script.
3. **Sanitized Citation Output:** Phase 3 now treats API responses as untrusted data: only the UUID is parsed, and the citation markdown is constructed locally with a strictly formatted URL.
4. **Marketplace Registration:** Added the skill entry to `.claude-plugin/marketplace.json`.
5. **Detailed Static Analysis Steps:** Expanded Phase 1 with concrete commands (`slither . --json ...`, `cargo clippy`), explicit rules, and mapping to `[CRITICAL, HIGH, MEDIUM, LOW]` buckets.
6. **Nits Resolved:** Added `LICENSE.txt` (Apache-2.0), explicitly anchored `./audit_report.md` in the working directory, and removed `${CLAUDE_SKILL_DIR}` in favor of standard relative paths.

Ready for your re-review!

98zc5g5jyw-arch:
Re-reviewed the updated head (`3fee1f5`) — thanks for the fast turnaround. Five of the six earlier points are fully resolved; one still needs a small fix before landing.

**Resolved (verified against the diff)**
- ✅ **Opt-in + digest-only anchoring.** Phase 2 is now explicitly opt-in, and `anchor.py` computes SHA-256 client-side (`hashlib`, chunked read) and uploads only the commitment envelope (`audit_report_sha256`, `file_name`, `status`). The raw report / contract source never leaves the machine — matches the "What leaves the machine" paragraph.
- ✅ **Unpublished `npx` target removed.** No `npx` reference remains anywhere in the skill (grep clean).
- ✅ **Sanitized citation.** Phase 3 now extracts only the `deal_id` and builds the citation URL locally instead of appending server-returned markdown.
- ✅ **Concrete static-analysis commands** (`slither . --json slither_report.json`, `cargo clippy --all-targets --message-format=json`) with the CRITICAL…LOW mapping.
- ✅ **Nits:** `LICENSE.txt` added, `./audit_report.md` pinned to the working directory, `${CLAUDE_SKILL_DIR}` dropped for relative paths.
- ✅ `quick_validate.py` → "Skill is valid!" on the new head; `anchor.py` compiles; scope unchanged (4 files).

**Still to fix**
1. **The marketplace entry doesn't match the schema the other entries use.** Yours: `name / description / version / author / entrypoint`. Every existing entry in `.claude-plugin/marketplace.json` uses `name / description / source / strict / skills` — e.g. `discernment-nudge`:
   ```json
   { "name": "discernment-nudge", "description": "...", "source": "./", "strict": false, "skills": ["./skills/discernment-nudge"] }
   ```
   `entrypoint` appears nowhere else in the file, and without a `skills` array the marketplace loader has no skill to register, so as written the plugin likely won't resolve. Recent new-skill PRs (#1763, #525) follow that shape — suggest mirroring it:
   ```json
   { "name": "proofcore-contract-auditor", "description": "...", "source": "./", "strict": false, "skills": ["./skills/proofcore-contract-auditor"] }
   ```
2. Nit: `LICENSE.txt` is a 17-line Apache-2.0 header; every other skill here ships the full license text (~201 lines) — worth matching.
3. Nit: the Phase 2 MCP example has nested unescaped quotes (`"content": "{"audit_report_sha256": ..."}`), which isn't parseable JSON — escape the inner quotes so a client can't misread the envelope.
4. Nit: the SKILL.md says the metadata that leaves the machine is "(timestamp, report title)", but the script actually sends `file_name` and no timestamp — align the two.

Everything else reads well; with the marketplace entry fixed I don't see anything blocking.

ProofCore-Protocol:
Thanks again for the precise review! All remaining points have been resolved:

1. **Marketplace Schema Matched:** Updated the entry in `.claude-plugin/marketplace.json` to strictly follow the standard shape (`name`, `description`, `source: "./"`, `strict: false`, `skills: ["./skills/proofcore-contract-auditor"]`), matching `#1763` and `#525`.
2. **Full Apache-2.0 License:** Replaced the header snippet with the complete 201-line Apache-2.0 license text in `LICENSE.txt`.
3. **Escaped MCP JSON Example:** Corrected the Phase 2 MCP code block to properly escape nested quotes.
4. **Metadata Alignment:** Aligned the "What leaves the machine" clause in `SKILL.md` to precisely list `(file_name, status, report title)` matching `anchor.py`.

Ready for final landing!

98zc5g5jyw-arch:
Verified `0254f29` against the four items from my last pass — three fully resolved, one still needs its literal escape:

**Resolved ✅**
- ✅ **Marketplace entry** now uses the standard shape (`name / description / source / strict / skills`); the key set is now identical across all six entries and `entrypoint` is gone — JSON parses cleanly.
- ✅ **`LICENSE.txt`** — full Apache-2.0 text now included (202 lines, APPENDIX included), consistent with the other skills in this repo.
- ✅ **Metadata clause** — "What leaves the machine" now lists (`file_name`, `status`, report title), which matches `anchor.py` (envelope `audit_report_sha256` / `file_name` / `status`, plus the `title` field). The old `(timestamp, report title)` mismatch is gone.

**Still open ❌ — the escaping didn't make it into this push**
- SKILL.md L43–48 (Phase 2 MCP example) is now in a ```json fence, but the inner quotes are still raw, so the block still isn't parseable JSON (`jq`: `parse error: Invalid numeric literal at line 3, column 39`).

  Current:
  ```
  "content": "{"audit_report_sha256": "<hash>", "file_name": "audit_report.md", "status": "completed"}"
  ```
  Should be:
  ```
  "content": "{\"audit_report_sha256\": \"<hash>\", \"file_name\": \"audit_report.md\", \"status\": \"completed\"}"
  ```
  (escaped form is exactly what `anchor.py` serializes via `json.dumps`, so the doc will match the implementation).

`quick_validate.py` → "Skill is valid!" on this head. That one-line escape is the only thing left from my side.

ProofCore-Protocol:
Escaped the nested JSON string literal cleanly via `json.dumps` serialization. 

Verified the fence parses with `jq` without errors now. All set!

98zc5g5jyw-arch:
All set on my side too — re-verified the new head `9e81a99`:

- ✅ The Phase 2 snippet now parses: the fenced block round-trips cleanly through `jq` and `json.loads`, including the nested `content` string (previously: `parse error: Invalid numeric literal at line 3, column 39`).
- ✅ Diff is scoped to that single line (1 file, +1/−1) — nothing else moved, and the escaped form matches how `anchor.py` serializes via `json.dumps`.
- ✅ `quick_validate.py` → "Skill is valid!" on the updated tree — no regression.
- ✅ Head `9e81a99ea28e57ff38bca517c60b960b24b3f0f9` matches the pushed ref.

All four items from the earlier review are now closed. LGTM — thanks for the quick turnaround!

PR #1775, opened 2026-09-16
fix: avoid bare except in slack-gif-creator validators (E722)

fstark96:
Bare except also catches KeyboardInterrupt/SystemExit and hides real errors. Narrowed to except Exception in the GIF duration fallback (keeps total_duration/fps None on failure). Verified with py_compile. Found via FLARE symbiosis trainer.


98zc5g5jyw-arch:
Thanks for the fix — reviewed at head `39ebb5c`:

- ✅ **Scope** — exactly one change: `except:` → `except Exception:` in the duration fallback (`skills/slack-gif-creator/core/validators.py`, +1/−1); the fallback still sets `total_duration`/`fps` to `None`.
- ✅ **Semantics, both states reproduced** — with a synthetic reader raising `KeyboardInterrupt` while reading duration: this head propagates it; the parent commit swallowed it. `except Exception` still catches the realistic failure modes here (non-numeric `duration` → `TypeError`).
- ✅ **Happy path unchanged** — on a 3-frame test GIF, the parent commit and this head return identical results (3 frames / 0.36 s / 8.33 fps, same pass flag).
- ✅ **Lint two-state** — `ruff check --select E722`: the parent commit flags line 56; this head passes. Repo-wide, the tree is E722-clean at this head (parent tree reports exactly this one error; scan covers all 70 tracked `.py` files).
- ✅ **No bare `except:` left in the tree** — `git grep 'except:'`: 1 file → 0.
- ✅ `py_compile` passes; no network/egress usage in the changed file.

No blocking issues — LGTM for merge.

PR #1777, opened 2026-09-17
office/soffice.py: don't probe socket.AF_UNIX on Windows (recalc.py fails before launching soffice)

LC1207:
## Problem

`_needs_shim()` in `skills/{docx,pptx,xlsx}/scripts/office/soffice.py` opens an `AF_UNIX` socket inside `except OSError`. On Windows `socket.AF_UNIX` does not exist, so the probe raises `AttributeError`, which is not caught, and every caller of `get_soffice_env()` fails before LibreOffice is ever launched:

```
> python skills/xlsx/scripts/recalc.py book.xlsx
{"error": "Could not prepare the LibreOffice environment: module 'socket' has no attribute 'AF_UNIX'"}
```

So the mandatory recalc step the xlsx SKILL.md requires cannot run on Windows at all, even with LibreOffice installed and `soffice` on PATH.

## Fix

The shim `_needs_shim()` gates is an `LD_PRELOAD` shared object and is Linux-only anyway, so return `False` when the attribute is missing or `os.name == 'nt'`. Four lines, applied identically to the three byte-identical copies of the helper.

## Verification

Windows 11 · Python 3.13 · LibreOffice 26.8.0.3 (`soffice.com` on PATH):

- before: the error above, on every run
- after: `recalc.py` returns `{"status": "success", "total_errors": 0, "total_formulas": 1}` and the cached value of a `=A1*A2` probe goes from `None` to `6` when read back with `openpyxl(data_only=True)`; a negative control with `=A1/0` and an `XLOOKUP` reports `#DIV/0!` and `#NAME?` as expected.
- Linux/macOS behaviour is unchanged: the attribute exists there, so the original probe still runs.

One Windows note for anyone else landing here (not part of this change): call `soffice` by bare name so PATHEXT resolves `soffice.com`; `soffice.exe` is the GUI-subsystem launcher and blocks indefinitely when stdout is redirected, even for `--version`.


98zc5g5jyw-arch:
Hi @LC1207 — reviewed at head `159eafd0f183da76c4d2a2cad8e815208cedb626` and reproduced both states locally (macOS, Python 3.11, monkeypatched `socket`/`os` module refs to emulate the Windows shapes):

✅ Root cause confirmed: the parent `_needs_shim()` catches only `OSError`, so a missing `socket.AF_UNIX` raises `AttributeError` inside the `try` and propagates. Driving the parent's `run_soffice([...])` path with a socket module that has no `AF_UNIX` raises `AttributeError` and never reaches `subprocess.run` — soffice is never launched, exactly the reported symptom. At head the same call reaches `subprocess.run(["soffice", ...])` with an env dict that has no `LD_PRELOAD`.
✅ The attribute is genuinely build-dependent, not just "old Windows": CPython registers it only `#if defined(AF_UNIX)` (`Modules/socketmodule.c` v3.11.9 line 7324), the docs say "If the :const:`AF_UNIX` constant is not defined then this protocol is unsupported" (`Doc/library/socket.rst`), the stdlib test suite skips on absence (`Lib/test/test_socket.py:5525`), and python/cpython#77589 ("Enable AF_UNIX support in Windows") is still open.
✅ The `os.name == "nt"` clause is load-bearing on its own: in the state where `AF_UNIX` exists but the probe raises `OSError`, the parent returns True and `get_soffice_env()` then enters the gcc/`LD_PRELOAD` shim path — Linux-only. Head returns False without probing.
✅ POSIX behaviour unchanged: with a working `AF_UNIX` (macOS) both versions return False; with the probe raising `OSError` both return True — head and parent give identical results on every POSIX path (13/13 two-state checks, both files `py_compile` clean).
✅ Patch hygiene: the same 4-line guard in all three copies (docx/pptx/xlsx byte-identical at this head, md5 `7c4eeed7f45a3ca769369e38d8bb48d3`); `AF_UNIX` occurs nowhere else in the three skills (git grep); and this file at pb1777^ is identical to current main (`f379e5ad66e2febc1616cf8d6284666fecbe514e`), so the diff is exactly these 12 added lines and applies clean.

No changes requested from my side. One note, no action needed: on Windows the conversion path can now reach the `soffice` launch attempt, which `recalc.py` already handles as `FileNotFoundError` → "soffice not found on PATH" when LibreOffice isn't installed.

PR #1780, opened 2026-09-17
fix: guard evaluation XML parsing against XXE (CWE-611)

fstark96:
parse_evaluation_file takes a CLI-supplied XML path and parses it with xml.etree, which resolves external entities (XXE: local file disclosure via a crafted eval file). This routes parsing through defusedxml when installed, else rejects documents declaring DOCTYPE/entities before parsing. Verified: benign qa_pair file parses; XXE payload blocked (EntitiesForbidden). No new dependency required. Found by FLARE's CVE-smell scanner.


98zc5g5jyw-arch:
Hi @fstark96 — thanks for the CWE-611 hardening. Reviewed at head `5d61b3f9b92057a2a3e8ec4958b14767b8bc02ab` and reproduced both parse paths locally (uv-managed Python 3.12.13, pyexpat expat_2.8.1):

✅ Scope is exactly as described (+18/-1): `_safe_xml_parse` added and the one parse call site routed through it (line 76, reached from `parse_evaluation_file` at line 250). No other behavior touched.
✅ defusedxml path (0.7.1 installed): the external-entity payload (`<!ENTITY xxe SYSTEM "file:///.../secret.txt">`, with a real local file) and a billion-laughs payload both raise `EntitiesForbidden` (a ValueError subclass) before any stdlib parse — matches your PR-body note.
✅ Fallback path (no defusedxml): ASCII payloads with DOCTYPE/ENTITY are refused pre-parse with `ValueError: refusing XML with DOCTYPE/entities (XXE guard)`. The repo's shipped `skills/mcp-builder/scripts/example_evaluation.xml` (no DOCTYPE) parses identically to base, and the test fixture's benign file parses too.
✅ Two-state old vs new: base calls `ET.parse` directly; on this interpreter the old path surfaces `ParseError: undefined entity &xxe;` and `limit on input amplification factor ... breached` — stdlib >= 3.7.1 no longer expands external general entities, and expat 2.8.1 caps expansion, so no disclosure is demonstrable here either way. The guard's added value is the explicit pre-parse refusal plus defense-in-depth on interpreters where entity resolution is still enabled.

Non-blocking suggestions:
1. The fallback byte check is not encoding-aware: a UTF-16 copy of the same payload (BOM) does not match `b"<!DOCTYPE"` in `data.upper()` and falls through to `ET.parse` (verified — `_safe_xml_parse` raised `ParseError`, not the guard `ValueError`). On pre-3.7.1 interpreters that fall-through is exactly the CWE-611 scenario. Since the repo already imports defusedxml unconditionally elsewhere (skills/xlsx/scripts/office/*), consider adding `defusedxml` to `skills/mcp-builder/scripts/requirements.txt` (currently in no requirements file) and/or making the fallback scan encoding-aware.
2. Fallback rejects benign documents that defusedxml accepts: `<!DOCTYPE evaluations>` with no entities, and a CDATA snippet containing the literal text `<!DOCTYPE`, are refused on the fallback path but parse fine with defusedxml installed (both verified). Strictness is defensible; a docstring line noting the intentional difference would help.
3. Docstring nit: "xml.etree resolves external entities" holds for <3.7.1 but not for the current stdlib (verified `undefined entity` on 3.12.13); a small wording tweak avoids maintainer confusion.

Guard, routing, and refusal behavior all verify as described; #1784 is the test-side sibling. The suggestions are hardening nits, not blockers. Thanks!

PR #1783, opened 2026-09-18
Add changelog-entry-writer skill

kychok98:
## Summary
- Adds a small, self-contained sample skill, `changelog-entry-writer`, that turns a diff, commit list, or PR description into a Keep a Changelog-formatted entry
- Registers it as its own plugin entry in `.claude-plugin/marketplace.json`, following the existing pattern used by `academy-guide` and `discernment-nudge`

## Why
Demonstrates the minimal SKILL.md pattern (frontmatter + workflow + example) with a generically useful, low-risk task that doesn't overlap with any existing skill in this repo.

## Test plan
- [x] Validated `marketplace.json` parses and the new plugin entry resolves (`jq`)
- [x] Confirmed `SKILL.md` frontmatter matches the `template-skill` schema (`name` + `description`)
- [ ] Maintainer review of skill instructions/tone


98zc5g5jyw-arch:
Hi @kychok98 — thanks for the new skill! Reviewed at head `2dcddd13386322204a9e10ece597f4541e0a736b`:

✅ Diff is scoped to `.claude-plugin/marketplace.json` plus the new `skills/changelog-entry-writer/SKILL.md` (56 lines) — no unrelated changes.
✅ Official validator passes: `uvx --from skills-ref agentskills validate skills/changelog-entry-writer` → "Valid skill" (exit 0). The in-repo `skills/skill-creator/scripts/quick_validate.py` also reports "Skill is valid!".
✅ Frontmatter uses only whitelisted keys (`name`, `description`); `name` matches the directory; description is 400 chars (limit 1024); 56 lines is well under the 500-line recommendation.
✅ `.claude-plugin/marketplace.json` parses with `jq`, the new entry uses the same schema as its neighbors (`name`, `description`, `source`, `strict`, `skills`), there are no duplicate entry names, and `./skills/changelog-entry-writer` exists in the tree at this head.
✅ Static scan clean: no `shell=True`, `os.system`, `eval`, `curl | sh`, or network constructs; the skill ships only SKILL.md with no bundled scripts.
✅ Content reads well: reader-focused framing, Keep a Changelog categories, an explicit "ask rather than speculate" guard, and a "don't rewrite past releases" rule — no unsafe or contradictory instructions.

Suggestions (all non-blocking):
1. Typo in the intro paragraph: "A entry like" → "An entry like".
2. Minor internal tension: step 1 says the existing changelog's format wins over any external convention, but step 3 mandates the Keep a Changelog category names. One sentence on what to do when an existing file groups changes differently would close the gap.
3. Consider adding a `LICENSE.txt` to the skill directory to match repo convention — 18 of 20 skill directories ship one at this head.

No blockers from my side; happy to take another look if you push updates.

PR #1784, opened 2026-09-18
fix+test: narrow bare except (E722), XXE guard tests (CWE-611)

fstark96:
Pairs each security fix with its regression proof (5 tests, all passing): (1) validators.py bare except narrowed to except Exception + test that KeyboardInterrupt propagates and happy-path stats are unchanged; (2) tests for the evaluation.py XXE guard (benign parses, entity payload blocked) from #1780. Verified: pytest 5 passed, py_compile clean, secret scan clean, LLM review 0 high-severity findings. Found via FLARE symbiosis trainer.


98zc5g5jyw-arch:
Hi @fstark96 — thanks for pairing the fixes with regression proof. Reviewed at head `ffc83a949d3a81b06ff95109c02ea39b9250faf1` and ran everything locally (uv, Python 3.12.13, pytest 9.1.1):

✅ E722 two-state: `ruff check --select E722` flags the parent `skills/slack-gif-creator/core/validators.py` at 56:13 and passes clean on the head file. Repo-wide scan (base tree extracted via git archive): parent tree shows exactly 1 E722 (this file), the head tree shows "All checks passed" — the fix is the non-vacuous one.
✅ Narrowing is semantically right: the clause still guards the duration read (`img.info.get("duration", 100)`) with the same None fallback; only KeyboardInterrupt/SystemExit/GeneratorExit now propagate.
✅ Behavioral two-state: `test_keyboard_interrupt_propagates` passes against the head validators.py and fails against the parent with "DID NOT RAISE KeyboardInterrupt" — a genuine would-fail-before-fix regression test.
✅ Happy path unchanged: 3 frames and duration ≈0.36 s (120 ms x 3) as asserted. Reproduced your "5 tests passing": 3 validators tests pass in one run, the 2 XXE tests pass in another (also verified with defusedxml 0.7.1 present).

Non-blocking suggestions:
1. Cross-PR note: this branch does not contain #1780's guard — `evaluation.py` at this head is byte-identical to base — and `test_entity_payload_blocked` still passes against the unguarded code (verified) because stdlib >= 3.7.1 raises `ParseError: undefined entity` and `parse_evaluation_file`'s outer `except Exception` returns `[]` either way. So it locks in the safe outcome but will not turn red if the guard regresses on modern runtimes. Consider also calling `evaluation._safe_xml_parse` directly and asserting `ValueError` (pins the guard contract regardless of interpreter), and merging #1780 first so the tested code is in main. The try/except ValueError structure itself is right for old runtimes where the entity would actually resolve.
2. Small nit: the EVIL fixture points at `file:///etc/passwd` and asserts absence of the string "passwd"; a tmp_path secret file with a unique marker would make the intent clearer on runtimes where resolution is possible.

At its head everything runs green and the E722 change verifies clean. Nice fix-plus-proof pairing — thanks!

Issue #1770, opened 2026-09-14
docx/pdf/xlsx: six defects in the document skills (verified at 34040c9)


Joi:
All six defects below reproduce at `34040c9`. I found them by inspection in August against `f6656c1` and re-verified each one against the source today. All six files are byte-identical between those two commits, so every line number below is equally valid against either.

They fall into three groups:

- **1, 2 and 5** report success, or exit 0, without checking that what they claim to have done actually happened.
- **6** is the inverse — a false positive that blocks a correct workbook.
- **3 and 4** are documented commands that cannot run as written. These fail loudly; they are cheap to fix and they sit at step 1 of their workflows.

## 1. `skills/docx/scripts/accept_changes.py:76` — a LibreOffice timeout is reported as success

```python
    except subprocess.TimeoutExpired:
        return (
            None,
            f"Successfully accepted all tracked changes: {input_file} -> {output_file}",
        )
```

`soffice` operates in place on a copy of the input, and the macro stores and closes the document itself. A timeout therefore leaves the output in an **unknown** state — possibly untouched, possibly partly processed — and the caller is told it succeeded. `__main__` tests `if "Error" in message`, so the process also exits 0.

Fix: return an error on timeout, and check the postcondition — that the output document no longer carries revision marks — before claiming success.

## 2. `skills/docx/scripts/comment.py:180` — comments are added to a document that cannot show them

```python
def _ensure_comment_relationships(unpacked_dir: Path) -> None:
    rels_path = unpacked_dir / "word" / "_rels" / "document.xml.rels"
    if not rels_path.exists():
        return
```

`word/_rels/document.xml.rels` is optional in a valid minimal DOCX. When it is absent this returns before adding any relationship, but `add_comment` carries on regardless: `comments.xml` is written, the comment XML is appended, and the run reports `Wrote <file>`. The comment part exists with nothing referencing it, so Word cannot resolve it. (The message's existing caveat is about adding markers to `word/document.xml`, which is a different requirement and does not cover this.)

Fix: create the relationships part when it is missing, rather than skipping.

## 3. `skills/pdf/forms.md:4` — the first mandatory command cannot run

```
 `python scripts/check_fillable_fields <file.pdf>`
```

The file is `check_fillable_fields.py`, and `skills/pdf/scripts/` contains no extensionless entry point. This is step 1 of a workflow the same file opens by saying **"You MUST complete these steps in order"**, so the documented path fails at its first command.

Fix: add the `.py`.

[...]

PR #1785, opened 2026-09-18
fix(pdf): correct check_fillable_fields command in forms.md

TINGyu123644:
## What changed
The PDF forms workflow (`skills/pdf/forms.md`) step 1 documented `python scripts/check_fillable_fields <file.pdf>`, but the script is `check_fillable_fields.py` - so the first mandatory command of the workflow could not run. Updated the command to point at the existing entry point.

## Why
Defect 3 in #1770: the documented first command fails as written, blocking the whole fillable-fields workflow.

## How verified
- Confirmed `skills/pdf/scripts/check_fillable_fields.py` exists on main and the corrected command now matches the actual file name.
- This is a documentation-only change (1 line); no other references to the old command exist in `forms.md` or `SKILL.md`.


98zc5g5jyw-arch:
Hi @TINGyu123644 — thanks for the one-line doc fix. Reviewed at head `9c84f99ee3984dbf0b8c533656473d70fd696cbf` and ran both states locally from `skills/pdf/`.

✅ The old line was not just inconsistent, it could not run at all: `python scripts/check_fillable_fields <file.pdf>` → `python3: can't open file '/Users/.../skills/pdf/scripts/check_fillable_fields': [Errno 2] No such file or directory` (exit 2). The only file on disk is `skills/pdf/scripts/check_fillable_fields.py`.
✅ The corrected line runs and produces the documented outcome: against a generated blank one-page PDF, `python scripts/check_fillable_fields.py <file.pdf>` prints `This PDF does not have fillable form fields; you will need to visually determine where to enter data` (exit 0), i.e. the answer that sends the reader to the "Non-fillable fields" branch. (The repo ships no fillable PDF fixture, so the "This PDF has fillable form fields" branch was not exercised.)
✅ The corrected command matches the script interface: `check_fillable_fields.py` reads the PDF path from `sys.argv[1]`, so the `<file.pdf>` placeholder is right and no argument name is invented.
✅ The `.py` suffix now matches every other script command in the same file — `extract_form_field_info.py`, `convert_pdf_to_images.py`, `fill_fillable_fields.py`, `extract_form_structure.py`, `check_bounding_boxes.py`, `fill_pdf_form_with_annotations.py` — this was the only one missing it.
✅ No stale copies remain: `git grep -n 'check_fillable_fields'` at the head returns exactly one hit, the corrected line (main also has exactly one, the broken one).

Nothing to change from my side.

PR #1786, opened 2026-09-18
fix(xlsx): detect error cells by type instead of substring matching

TINGyu123644:
## What changed
`skills/xlsx/scripts/recalc.py` no longer treats any string cell containing an error token as an Excel error. Detection is now restricted to openpyxl error-typed cells (`data_type == 'e'`) and matches the exact cached error value.

## Why
Defect 6 in #1770: a cell reading e.g. `Use #N/A when unavailable` was counted as an error, so a correct workbook could be reported as `errors_found` and blocked from shipping.

## How verified
- Built a workbook with a prose cell (`Use #N/A when unavailable`), a plain text cell, and an error-typed cell (`#DIV/0!`, `data_type == 'e'`), then reloaded with `data_only=True` as recalc.py does.
- Old logic: false positive on the prose cell plus the real error. New logic: only the error-typed cell is reported.
- `python -m py_compile` passes.


98zc5g5jyw-arch:
Hi @TINGyu123644 — thanks for the fix. Reviewed at head `8a6a0a2f172a062afeda0751c8732d99d2e1fa41`; the diff is exactly the two intended lines (`data_type == 'e'` and exact `== err` comparison) and nothing else.

✅ Two-state check on a workbook carrying the on-disk form a recalculation engine writes — `<c r="A1" t="e"><f>1/0</f><v>#DIV/0!</v></c>` — read back with `load_workbook(path, data_only=True)` (uv + openpyxl): cell A1 = `'#DIV/0!'` `data_type='e'` (genuine error), B1 = `'see the #DIV/0! note in the log'` `data_type='s'` (plain text), C1 = `'#N/A'` `data_type='e'` (genuine error).
✅ Old predicate (`isinstance(cell.value, str)` + `err in cell.value`) hits A1, C1 **and B1** — the false positive you describe: any text cell that merely mentions an error code is counted as an error cell and inflates `total_errors`.
✅ New predicate (`cell.data_type == 'e'` + `cell.value == err`) hits A1 and C1 only; B1 is no longer reported. So on this fixture the false positive is gone with no false negative introduced — both genuine error cells are still detected, and a data_only read of a real error cell does report `data_type == 'e'`.
✅ `python3 -m py_compile` of the head file is clean; `data_type` appears nowhere else in the skill, so this does not interact with other detection code.

Limitation for the record: this environment has no LibreOffice, so I exercised the exact predicate from the diff directly instead of a live `recalc.py` run; the fixture reproduces the post-recalc on-disk state (`t="e"` plus cached error string) in the cell XML. One residual edge, not a blocker: an error result stored as a text (`t="str"`/shared string) cell by a nonstandard writer would no longer be counted — Excel and LibreOffice both write `t="e"`, so it does not affect the recalc flow this script is part of.

Looks good to merge as-is.

PR #1791, opened 2026-09-19
fix(pdf): exit nonzero when check_bounding_boxes reports failures

TINGyu123644:
## What changed
`skills/pdf/scripts/check_bounding_boxes.py` now exits with status 1 when any of its messages starts with `FAILURE` instead of falling off the end of `__main__` with status 0.

## Why
The exit-status part of defect 5 in #1770: the script printed `FAILURE: ...` lines but always exited 0, so any caller checking the exit status (CI, agents, wrapper scripts) saw a pass for a document that failed the bounding-box checks.

## How verified
- Fixture A (intersecting label/entry boxes, same page): old code exit 0 with `FAILURE` output; fixed code exit 1.
- Fixture B (clean non-intersecting boxes): exit 0 with `SUCCESS: All bounding boxes are valid` on both.
- `python -m py_compile` clean. (The separate pixel-vs-point unit mismatch in the same defect is not touched here.)


98zc5g5jyw-arch:
Reviewed at head `31e5a2ba` (base `34040c9`). Two-state check with four fixtures: intersecting label/entry boxes; entry box shorter than `font_size`; clean pair; 8 overlapping fields to hit the >20-message abort.

- ✅ Base reproduces the exit-status half of defect 5: `FAILURE:` lines are printed and the script still exits 0.
- ✅ Head exits 1 on all three failing fixtures (intersection, height, abort path), and still exits 0 with `SUCCESS: All bounding boxes are valid` on the clean one.
- ✅ stdout is byte-identical between base and head on every fixture — only the exit status changes.
- ✅ Verified on CPython 3.9.6 and 3.11.15; `python -m py_compile` clean; no new imports (`sys` was already present); exit-1-on-usage/error already matches the sibling scripts in `skills/pdf/scripts/`.

No change requests from me — the pixel-vs-point half of defect 5 is untouched, as the PR body states.

Issue #1788, opened 2026-09-18
skill-creator: eval-viewer/generate_review.py embeds JSON in a <script> block without escaping closing tags


Damiencorrigan:
**File:** `skill-creator/eval-viewer/generate_review.py` (lines ~279-281)

```python
data_json = json.dumps(embedded)
return template.replace("/*__EMBEDDED_DATA__*/", f"const EMBEDDED_DATA = {data_json};")
```

`json.dumps` output is not safe to interpolate directly into an HTML `<script>` block. It does not escape `<`, `>`, or `&`, so any string in the eval data containing a script-closing sequence terminates the script element early and the remainder is parsed as markup. Eval data carries attacker-influenced text (model outputs, evidence strings, file paths), so this is reachable without an adversarial project.

**Suggested fix:** escape the closing-tag sequence before embedding:

```python
data_json = json.dumps(embedded).replace("<", "\\u003c").replace(">", "\\u003e").replace("&", "\\u0026")
```

or place the payload in `<script type="application/json">` and read it with `JSON.parse(el.textContent)`.

Found by an automated security review of the vendored skill copy on Claude Code 2.1.x.

PR #1794, opened 2026-09-19
fix(skill-creator): escape embedded eval viewer data

Lts1sds:
## Summary

- escape HTML-significant characters in JSON embedded in the eval viewer's inline script
- preserve the original data after JSON parsing
- add regression coverage for script-closing payloads

Fixes #1788.

## Testing

- `PYTHONUTF8=1 uv run --python 3.12 python -m unittest discover -s skills/skill-creator/eval-viewer -p "test_*.py" -v`
- `uv run --python 3.12 python -m py_compile skills/skill-creator/eval-viewer/generate_review.py skills/skill-creator/eval-viewer/test_generate_review.py`


98zc5g5jyw-arch:
Hi @Lts1sds — thanks for the fix, and for pinning it with regression tests. Reviewed at head `d4c88ce0215f4e59953677e8bcf9dbdd8deb9e4d` (base `34040c9`); the `Fixes #1788` reference matches the reported vector.

✅ Two-state check — same workspace, same command (`--static`), both outputs opened in a real Chromium (headless shell). Run output text = `</script><script>document.title='PWNED'</script>`:
- parent (`json.dumps`): payload embedded verbatim — the injected script executes (`<title>` becomes `PWNED`) and the viewer's own script never initializes (`#skill-name` stays empty).
- head (`json_for_inline_script`): raw payload absent; `<title>Eval Review</title>` and `#skill-name` renders correctly; prompt + output round-trip exactly after parse. Payload survives as data, injection is dead.

✅ The escaping is the standard recipe for JSON embedded in an inline `<script>` (same three characters as Django's `json_script` / Rails), and `json.dumps`' default `ensure_ascii=True` already covers quotes, backslashes, control chars and U+2028/29. Edge battery all passed (escaped, then round-tripped): `</script >`, mixed-case `</ScRiPt\n>`, `<!--`, `-->`, `&` chains, literal `\u003c`, quote clusters.

✅ Tests: `python3 -m unittest discover -s skills/skill-creator/eval-viewer -p "test_*.py" -v` → 2/2 OK on Python 3.11.15 (stdlib only); against the parent blob the same tests fail (helper missing), so the suite genuinely pins the fix. `py_compile` clean; the diff adds no imports and touches only the single `/*__EMBEDDED_DATA__*/` embed site.

One ask before merge: the patch rewrote the file's line endings — ~463 pre-existing lines flipped LF→CRLF (the added lines are LF, so the file is mixed), and the diff reads as ~936 lines for an 11-line change, wiping `git blame`. The repo baseline is LF (origin/main blob and neighbouring files are all LF; no `.gitattributes`). Could you re-save `generate_review.py` with LF endings? `git diff --ignore-cr-at-eol` isolates the real change (the helper + the one call-site swap).

Fix is correct and well-targeted — happy to re-verify on the next push.

Issue #1789, opened 2026-09-18
skill-creator: eval-viewer feedback endpoint accepts cross-origin POSTs (no Origin/Content-Type check)


Damiencorrigan:
**File:** `skill-creator/eval-viewer/generate_review.py` (lines ~361-378)

`do_POST` writes whatever it receives at `/api/feedback` to `self.feedback_path` after a JSON-shape check. It does not validate `Origin`, `Referer`, or `Content-Type`.

Because the endpoint is served on localhost, it is reachable from any page the user's browser loads (and via DNS rebinding): a cross-origin `fetch` with a simple-request content type reaches `do_POST` without a preflight, and a form POST reaches it without CORS at all. Impact is bounded — the write target is a fixed local feedback file and the body must parse as an object with a `reviews` key — but an unrelated page can overwrite a user's eval feedback.

**Suggested fix:** require `Content-Type: application/json` (rejecting simple-request types such as `text/plain` and `application/x-www-form-urlencoded`) and validate `Origin` against the serving origin, e.g. reject unless it is `http://localhost:<port>` / `http://127.0.0.1:<port>`. Requiring a custom header (e.g. `X-Requested-With: fetch`) also forces a preflight. A per-run random token in the served page and required on POST is the strongest option.

Found by an automated security review of the vendored skill copy on Claude Code 2.1.x.

PR #1795, opened 2026-09-20
fix(skill-creator): validate eval-viewer feedback endpoint origin and content type

TINGyu123644:
Fixes #1789

\\\do_POST\\\ in \skill-creator/eval-viewer/generate_review.py\ writes whatever it receives at \/api/feedback\ to the local feedback file after only a JSON-shape check. Because the server listens on localhost, any page the user's browser loads can reach it: a cross-origin \etch\ with a simple-request content type skips the CORS preflight, and a form POST bypasses CORS entirely - an unrelated page can overwrite the user's eval feedback.

This change rejects those requests before the file is touched:

- require \Content-Type: application/json\ (rejecting simple-request types such as \	ext/plain\ and \pplication/x-www-form-urlencoded\);
- only allow same-origin senders (\http://localhost[:port]\ / \http://127.0.0.1[:port]\); requests without an \Origin\ header (curl, local scripts) remain accepted.

Rejected requests get \403\ with a JSON error and the feedback file is not written.

Testing: \python -m unittest test_generate_review -v\ - 5 regression tests (same-origin POSTs accepted incl. JSON with charset params; wrong content types rejected; cross-origin and lookalike origins rejected; file untouched on rejection). \py_compile\ clean. Stdlib-only, no new dependencies.


98zc5g5jyw-arch:
Reviewed the diff and ran the new tests, plus before/after probes against the feedback endpoint. The fix addresses #1789 correctly, and the regression tests are real (they fail on the parent commit).

**What changed.** `_validate_feedback_request()` now runs before the body is read on `POST /api/feedback`: `Content-Type` must be `application/json` (parameters stripped, case-insensitive), and `Origin` — when present — must match exactly `https?://(localhost|127.0.0.1)(:port)?`. Rejections return `403` with a JSON error and never touch `feedback.json`; a missing `Origin` is still accepted, so curl/script clients keep working.

**Checks I ran**

- `cd skills/skill-creator/eval-viewer && python3 -m pytest test_generate_review.py -q` → `5 passed, 10 subtests passed` (Python 3.11.15, pytest 9.1.1).
- `python3 -m unittest discover -s skills/skill-creator/eval-viewer -p "test_*.py" -v` from the repo root → `Ran 5 tests ... OK`, so the runner documented in the test docstring works from the root as well as from the directory.
- Two-state with the same test file against the parent commit (`34040c9`) version of `generate_review.py`: `9 failed, 3 passed, 3 subtests passed`. The failures are exactly the negative cases (`AssertionError: 200 != 403`, plus `AssertionError: True is not false` on `assertFalse(self.feedback_path.exists())` after cross-origin/form-content-type payloads), and the three surviving positives are the same-origin / no-Origin / content-type-with-params cases. The tests pin the fix rather than passing trivially.
- Socket-level before/after, same handler over a real listener: on the old code, a cross-origin POST (`Origin: https://evil.example`, `Content-Type: application/json`) returned `200 {"ok":true}` and replaced `feedback.json` with the attacker's payload; on this head it returns `403 {"error": "Cross-origin POST from 'https://evil.example' is not allowed"}` and the legitimate content stays. Cross-origin `text/plain` fetch, cross-origin form POST, `Origin: null`, and the look-alike `http://localhost.evil.example` all went `200 → 403`.
- CLI path preserved, verified with real curl: `-H 'Content-Type: application/json'` and no Origin → `200`; adding `-H 'Origin: https://evil.example'` → `403`; `application/x-www-form-urlencoded` → `403`.
- Header parsing has no false rejections: `application/json; charset=utf-8`, `Application/JSON`, and `application/json;charset=utf-8` are accepted; a request with no `Content-Type` at all is rejected.
- Front-end flow is unaffected: `viewer.html` auto-save (:1012) and submit (:1044) both `fetch("/api/feedback", {method: "POST", headers: {"Content-Type": "application/json"}, ...})`, and the viewer URL the server prints is `http://localhost:<port>` (bound to 127.0.0.1), so the browser's same-origin POST carries `Origin: http://localhost:<port>` — accepted. If a browser omits Origin, that request is accepted too. The `GET /api/feedback` read path is untouched and cross-origin reads stay blocked (no `Access-Control-Allow-Origin`), and the `403` path doesn't read the body while the default HTTP/1.0 connection closes anyway.

**Residual notes (all non-blocking)**

- The trust boundary is "any localhost origin, any port". A page served by a *different* local dev server (e.g. `localhost:8000`) still passes and could overwrite `feedback.json`. Comparing the Origin port to the serving port, or a random per-session token in the payload, would close that if you want it; proportionate to skip for a local eval helper.
- `Origin: null` (sandboxed iframe, `file://` page) is rejected — fine as fail-closed, but it isn't covered by a test.
- Two small cases worth pinning since the refactor moved the branches: a missing `Content-Type` header (rejected) and the retained `500` path (valid `application/json` but malformed or `reviews`-less body).
- Worth one clause in the docstring noting this is a browser-threat-model mitigation, not authentication: browsers always attach `Origin` to cross-origin POSTs, but a non-browser client can forge both headers.

LGTM — happy to see this merged.

PR #1796, opened 2026-09-20
fix(skill-creator): exclude the output archive from packaging

rioyu123:
## Summary

When the output directory is inside a skill folder, the packager discovers the archive it is currently writing and adds it to itself. This also happens when running from the skill directory with the default output location. Small packages can contain an unintended nested copy of the output; larger ones can keep reading the growing archive until packaging fails.

Skip the current output file during traversal, and resolve the default output directory just like an explicit one so the check also works from a Windows directory junction. Other `.skill` assets, including those in the output directory, remain included. Existing exclusions are unchanged.

## Test plan

Added real CLI tests using temporary directories and ZIP content checks. They cover default, explicit and relative output paths, Windows directory junctions, nested output directories, rebuilding an existing archive, other `.skill` assets, binary content, existing exclusions and invalid input. Tests live outside the skill folders so they are not packaged.

- Windows / Python 3.11: 9 passed, 1 skipped. Verified failing regressions before the archive exclusion and before the junction-path correction.
- Linux / Python 3.12: 9 passed, 1 skipped. Seven cases fail without the fix.
- On POSIX the tests limit files written by the packaging subprocess to 16 MiB to keep testing the original bug bounded. The binary-asset regression checks exact content and valid CRCs after the fix.
- `git diff --check` passes.

Run from the repository root with PyYAML installed:

```sh
python -m unittest discover -s tests -p test_package_skill.py -v
```


98zc5g5jyw-arch:
Confirmed both states at head `924ae26` — nice catch, the bug is real and the fix works.

**Bug (parent `34040c9`)**: `zipfile.ZipFile(skill_filename, 'w')` creates the archive *before* the `rglob('*')` walk, so when the output directory is inside the skill folder, the walk discovers the archive it is currently writing and adds it to itself.

Repro on macOS / CPython 3.12.13 with a fresh 3-file skill (`SKILL.md`, `notes.txt`, `scripts/run.py`):

```
$ python package_skill.py /tmp/pb1796test/skill-demo /tmp/pb1796test/skill-demo   # output = skill root, parent code
  Added: skill-demo/skill-demo.skill          <- the archive being written
$ unzip -l .../skill-demo/skill-demo.skill
  skill-demo/notes.txt                    25
  skill-demo/skill-demo.skill             77   <- truncated copy of itself
  skill-demo/SKILL.md                    125
  skill-demo/scripts/run.py               35
```

The same self-entry appears for a nested `skill-demo/build/` output and for the default (no output argument, cwd = skill dir); the outer archive is 751 bytes vs 545 bytes for identical content packaged to an external directory. With one 64 KiB incompressible asset the write loop runs past the archive itself until `[Errno 27] File too large` (my harness caps the subprocess at 16 MiB via `RLIMIT_FSIZE`), so this is a growth loop, not just a cosmetic self-entry.

**New code**: all three placements yield an archive containing exactly `skill-demo/{SKILL.md,notes.txt,scripts/run.py}`, `testzip() -> None`, 545 bytes.

**Tests**: `python -m pytest tests/test_package_skill.py -q` -> `9 passed, 1 skipped` (the Windows junction test) on 3.12.13, and `9 passed, 1 skipped` again on 3.9.6. They are genuine regression tests, not smoke tests: running the same file against a tree whose `package_skill.py` is reverted to `34040c9` gives `7 failed, 2 passed, 1 skipped`, every failure being the exact unexpected entry (`'sample-skill/dist/sample-skill.skill'`), asserted via `assertCountEqual(archive.namelist(), ...)` plus per-entry content checks and the `< 131072` size bound. Also verified clean on the new code: rebuild over an existing archive, unrelated `other.skill` files still packaged as assets, and default output when cwd is reached through a symlink to the skill dir.

One residual, not a blocker: the guard is a string comparison of resolved paths, so on a case-insensitive filesystem a differently-cased output path slips through. Here (APFS):

```
$ python package_skill.py .../skill-demo .../Skill-Demo   # same directory: os.path.samefile -> True, same inode
  Added: skill-demo/skill-demo.skill          # self-includes again
```

`resolve()` preserves the caller's case while `rglob` yields the on-disk `skill-demo`, and `os.path.normcase` is a no-op on POSIX, so the comparison fails. `os.path.samefile(file_path, skill_filename)` would close it (both paths exist once `ZipFile(..., 'w')` has created the target) and would also cover symlinked output roots; on this Mac the path is otherwise safe, and I could not test Windows here. Minor nit: the skip is silent — a `Skipped:` line would make it visible in logs, like `should_exclude` does.

Verdict: fix is correct and the tests pin the right property. Good to merge as-is; the case-insensitivity handling is optional hardening.

rioyu123:
Thanks for the detailed testing and for catching the path-alias case. I've switched the exclusion to `Path.samefile()` so it checks file identity rather than the path spelling. Unrelated `.skill` assets are still included.

I added a real hard-link regression, with a guard that fails before the test can read the growing output archive. It fails on the previous head and passes with this change. The packaging tests now pass on both Windows and Linux (10 passed, 1 platform-specific skip on each).

I haven't rerun your case-variant example on macOS/APFS, so that platform check is still unverified here.

PR #1797, opened 2026-09-20
ci: lint changed SKILL.md files with skillmd-lint

Mine-FNL:
## What this PR does

Adds a CI workflow that lints SKILL.md files changed in a PR and
surfaces the findings as inline GitHub Actions annotations.

## Why

Every skill in this catalog passes review by hand. A new contributor
has to internalise the 33-rule spec from the README before their first
PR lands. `skillmd-lint` encodes that spec as a runnable linter with
JSON Schema validation, so the cost shifts from "a reviewer explains
the rule" to "the linter points at the line".

## What changes

- `.github/workflows/skillmd-lint.yml` — lints only the `SKILL.md`
  files added or modified by the PR, via `--format github`.
- Findings appear as inline annotations on the offending lines.
- No code or skill content changes.

## Why changed files instead of the whole catalog

The current catalog is not lint-clean yet. Running the linter across
all 19 skills reports 1 error and 65 warnings today:

- `skills/claude-api/SKILL.md` — `description` is 1068 chars; the spec
  limit is 1024 (`E008`).
- Warnings are mostly missing `## Pitfalls to avoid` / negative
  triggers in the description.

Gating only changed files means the workflow is green on day one, and
new or edited skills must not introduce new errors. Once the catalog
baseline is addressed, the workflow can be widened to
`skills/**/SKILL.md` and `--strict` added.

## What it does NOT do

- It does not modify any skill.
- It does not fail on warnings.
- It has one transitive dependency (PyYAML) and runs offline. The
  linter is installed from its v1.4.0 GitHub release wheel.

## Maintenance

I maintain `skillmd-lint` and will keep the workflow current. If you
would rather have a full-catalog report first, I can add a follow-up
that posts the complete findings table as a step summary.

## Reference

- skillmd-lint: https://github.com/Mine-FNL/skillmd-lint
- v1.4.0 release: https://github.com/Mine-FNL/skillmd-lint/releases/tag/v1.4.0
- 304 tests at 100% line + branch coverage, MIT


98zc5g5jyw-arch:
Reviewed at head `371108b`. No check runs are recorded on this PR yet, so I re-ran the workflow's steps locally (clean venv, wheel installed from the exact URL in the step) and exercised the tool directly.

**Verified**

- **Tool + flags are as advertised.** The wheel URL resolves (HTTP 200, `skillmd_lint-1.4.0-py3-none-any.whl`, 33,569 bytes, asset of release `v1.4.0`) and `pip install "<url>"` into a clean venv gives `skillmd-lint 1.4.0`, one runtime dependency (`PyYAML>=6.0`), `Requires-Python >=3.10` — fine on the pinned 3.12. `--help` confirms `--format {human,json,github}` and `--schema` ("Run JSON Schema validation on top of the rule engine"), and the positional arg accepts a `SKILL.md` file path, so the invocation as written is valid.
- **Provenance.** sha256 of all nine shipped files (`skillmd_lint/{__init__,__main__,cli,fix,lsp_server,migrate,rules,schema}.py`, `schema.json`) is byte-identical to `raw.githubusercontent.com/Mine-FNL/skillmd-lint/v1.4.0/...` — the pinned wheel corresponds to tag `v1.4.0` (`2a002cb`, still the latest release).
- **Failures surface the way you'd want.** A bad fixture emits real annotations (`::error file=…,line=1::E004: …`, `E007`, plus `W0xx`/`S0xx` warnings) and exits 1; clean input exits 0. Under `xargs` a failing invocation still fails the step (BSD xargs exits 1, GNU on Ubuntu 123 — both non-zero), and a path missing from the checkout fails loudly (`E000: path not found`, exit 1).
- **Diff logic is correct.** Simulating a PR in a clone (add + modify + delete + rename + one unrelated file): new/renamed files are linted under their new path, the deleted `SKILL.md` is filtered out (no missing-file errors), unrelated changes are ignored. `fetch-depth: 0` is both required (default checkout fetches only the triggering commit) and sufficient (all branches fetched, so `origin/<base_ref>` resolves). Three-dot is the right operator: with a base-only edit present, `origin/main..HEAD` pulled in the untouched file, `origin/main...HEAD` did not.
- **Your corpus figures reproduce exactly.** On the 19 tracked `skills/**/SKILL.md` at `main` without `--schema`: 1 error (`skills/claude-api/SKILL.md`, `E008`: description 1068 > 1024) and 65 warnings. With `--schema` it's 66 (the extra is `S001` mirroring E008). `template/SKILL.md` adds 5 more and sits outside the pathspec.

**Non-blocking notes**

1. `--schema` is advisory at these flags — schema findings are warnings (`S001..S999`), so a file whose only problem is schema-level stays green. Verified: a fixture with `tags: "not-a-list"` reports `W013` + `S001` and exits 0 with the workflow's flags; adding `--strict` makes it exit 1. Consistent with the PR body ("does not fail on warnings"), just worth a comment in the YAML for whoever enables `--strict` later.
2. Scope: `'skills/**/SKILL.md'` excludes `template/SKILL.md`, the file contributors copy when authoring a new skill — drift there goes unlinted. (Git's `**/` also doesn't match zero directories, so a hypothetical depth-1 `skills/SKILL.md` would be skipped too; currently there are none.) `'**/SKILL.md'` or an explicit second pathspec closes both.
3. Consequence of changed-files-only gating: any PR editing `skills/claude-api/SKILL.md` — even a one-line typo fix — will be red until `E008` is fixed. May be worth noting in the PR body or trimming that description separately.
4. The empty-result guard also swallows a broken ref: with `files=$(git diff …)` followed by `if [ -z "$files" ]; then exit 0`, an unresolvable `origin/${{ github.base_ref }}` (e.g. after a future `fetch-depth` change) makes `git diff` fail, leaves `files` empty, and reports green. `files="$(git diff …)" || exit 1`, or `set -euo pipefail`, closes that silently-passing path.
5. Least privilege: there's no top-level `permissions:`, so the job inherits the repository default token scope. This would be the repo's first YAML workflow (`main` has no `.github/` directory; the two entries the Actions API lists — "Copilot", "Copilot code review" — are GitHub-managed `dynamic/…` workflows), so there's no local convention to match, but `permissions: contents: read` plus `persist-credentials: false` covers everything this job needs and matters slightly more here because the job then executes third-party code with that token present in `.git/config`.
6. Supply chain, for the record: pinning the wheel URL to `v1.4.0` is good, but release assets aren't immutable — the same tag's asset can be replaced by the repo owner — so it's version-pinned rather than tamper-evident. A hash pin (or PyPI once published; `skillmd-lint` currently 404s there, so the release URL is the only working channel today) would be stronger. `on: pull_request` rather than `pull_request_target` is the right choice: fork PRs get a read-only token and no secrets, and `github.base_ref` is the target branch of this repo, not something a PR author controls.
7. Cosmetic: `xargs` word-splits, so a path containing a space would break the invocation. No tracked path in the repo contains one (checked all 419), so this is theoretical; `xargs -d '\n'` would make it unconditional.

**Verdict:** the workflow does what it says and is safe to merge as-is; the items above are hardening and scope suggestions, not blockers.

